PR #53, opened 2017-10-17
Code golf Stream._first_stage()

moreati:
This shaves a few more bytes off the code. There are a few more steps that could be taken, but I wasn't sure about. I'll add inline comments for each of those

### DIFF
--- a/mitogen/master.py
+++ b/mitogen/master.py
@@ -714,13 +714,12 @@ class Stream(mitogen.core.Stream):
             for f in R,r,W,w:os.close(f)
             os.environ['ARGV0']=e=sys.executable
             os.execv(e,['mitogen:CONTEXT_NAME'])
-        else:
-            os.write(1,'EC0\n')
-            C=zlib.decompress(sys.stdin.read(input()))
-            os.fdopen(W,'w',0).write(C)
-            os.fdopen(w,'w',0).write('%s\n'%len(C)+C)
-            os.write(1,'EC1\n')
-            sys.exit(0)
+        os.write(1,'EC0\n')
+        C=zlib.decompress(sys.stdin.read(input()))
+        os.fdopen(W,'w',0).write(C)
+        os.fdopen(w,'w',0).write('%s\n'%len(C)+C)
+        os.write(1,'EC1\n')
+        sys.exit(0)
 
     def get_boot_command(self):
         source = inspect.getsource(self._first_stage)
--- a/mitogen/master.py
+++ b/mitogen/master.py
@@ -715,11 +715,11 @@ class Stream(mitogen.core.Stream):
             os.environ['ARGV0']=e=sys.executable
             os.execv(e,['mitogen:CONTEXT_NAME'])
         else:
-            os.write(1, 'EC0\n')
-            C = zlib.decompress(sys.stdin.read(input()))
+            os.write(1,'EC0\n')
+            C=zlib.decompress(sys.stdin.read(input()))
             os.fdopen(W,'w',0).write(C)
             os.fdopen(w,'w',0).write('%s\n%s'%(len(C),C))
-            os.write(1, 'EC1\n')
+            os.write(1,'EC1\n')
             sys.exit(0)
 
     def get_boot_command(self):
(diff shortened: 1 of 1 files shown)


mitogen/master.py
```diff
@@ -706,21 +706,20 @@ def on_shutdown(self, broker):
...
-            os.fdopen(W,'w',0).write(C)
-            os.fdopen(W2,'w',0).write('%s\n%s' % (len(C),C))
-            os.write(1, 'EC1\n')
-            sys.exit(0)
+            os.dup2(r,101)
+            for f in R,r,W,w:os.close(f)
+            os.environ['ARGV0']=e=sys.executable
+            os.execv(e,['mitogen:CONTEXT_NAME'])
+        os.write(1,'EC0\n')
+        C=zlib.decompress(sys.stdin.read(input()))
```
moreati: Is this line correct for Python 3.x?

AIUI `input()` evaluates the first line as an integer, which is passed to `sys.stdin.read()` as the number of bytes to request. However on Python 3 `input()` does what `raw_input()` does on Python 2. So on Python 3 this line is passing a string as the length argument to `sys.stdin.read()`.

moreati: Also, could the use of `sys.stdin.read(...)` be replaced with `os.read(1, ...)`?

mitogen/master.py
```diff
@@ -706,21 +706,20 @@ def on_shutdown(self, broker):
...
+        os.fdopen(W,'w',0).write(C)
+        os.fdopen(w,'w',0).write('%s\n'%len(C)+C)
+        os.write(1,'EC1\n')
+        sys.exit(0)
```
moreati: Is this call to `sys.exit(0)` necessary. AIUI just reaching the end of the code, which we have, causes Python to exit with a status of 0.

mitogen/master.py
```diff
@@ -706,21 +706,20 @@ def on_shutdown(self, broker):
...
-            os.execv(sys.executable,['mitogen:CONTEXT_NAME'])
-        else:
-            os.write(1, 'EC0\n')
-            C = zlib.decompress(sys.stdin.read(input()))
-            os.fdopen(W,'w',0).write(C)
-            os.fdopen(W2,'w',0).write('%s\n%s' % (len(C),C))
-            os.write(1, 'EC1\n')
-            sys.exit(0)
+            os.dup2(r,101)
+            for f in R,r,W,w:os.close(f)
```
moreati: `str(map(os.close,(R,r,W,w)))` has the same character count, but would eliminate the `for` block. This would allow adding a step to `get_boot_command()` that munges the entire `if` block into a single line, with `;` delimiters, saving some `\n` bytes.

The `str()` wrapper would be necessary to 'pump' the `map` builtin on Python 3.

mitogen/master.py
```diff
@@ -706,21 +706,20 @@ def on_shutdown(self, broker):
...
+            os.environ['ARGV0']=e=sys.executable
```
moreati: If `sys,stdin` and `sys.exit` references can be eliminated, then this is the only remaining reference to `sys`. I can't find a substitute though.

moreati:
I won't be offended if you reject some, most or all of this PR. It was fun in itself. If you do want to go further down this road, then it might be good publicity/effective to submit this part of mitogen to /r/codegolf, or https://codegolf.stackexchange.com/.

dw:
Another fun contribution :)  I don't really mind this stuff to be honest, as long as it doesn't hurt readability too much. Actually I was thinking with your zlib change, the first stage doesn't need to be so ugly any more.

I think when it comes time to supporting Python 3 properly, everything is going to end up growing by 5-10% anyway, so in the long run changes like this might not be worth so much. Meanwhile they're fun and accepting them encourages further contribution, so it's fine :)

moreati:
Thank you for merging this. Did you read the comment about using `input()` on Python 3.x. Is it a bug?

dw:
Yes it's a bug. Honestly the Python 3.x port is going to be a nightmare, so little things like this don't concern me so much. I am much more concerned with compatibility back to Python 2.4 for now, and keeping the codebase compatible with 3 and that far back in the 2.x series will be a huge headache when the time comes to porting.

For example, 2.4 does not support "except foo as bar", while 3.x does not support "except foo, bar".

When the time comes, I'm thinking of experimenting with 2to3 or 3to2, but in the meantime all versions of 2.x are my target, as they're what's actually installed on the infrastructure I see while consulting, and will likely continue to see for another few years.

PR #54, opened 2017-10-20
tests: Fix GetModuleViaPkgutilTest fetching .pyc

moreati:
On my laptop (Ubuntu 17.10, Python 2.7.14 in a virtualenv), `test_regular_mod` fails with

```
AssertionError: "\nimport sys\n\n\ndef say_hi():\n    print 'hi'\n" !=
'\x03\xf3\r\n\xbbW\xd5Yc\x00\x00\x00\x00\x00\x00 ... /home/alex/src/mitogen/tests/data/module_finder_testmod/regular_mod.pyt\x08\x00\x00\x00<module>\x02\x00\x00\x00s\x02\x00\x00\x00\x0c\x03'
```

`__file__` contains the path of the compiled `.pyc`, not the `.py` source file that this test expects.

### DIFF
--- a/tests/module_finder_test.py
+++ b/tests/module_finder_test.py
@@ -1,5 +1,6 @@
-
+import inspect
 import unittest
+
 import mitogen.master
 
 import testlib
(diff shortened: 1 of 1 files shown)


dw:
The module server is due a third rewrite any day now. Some of the tricks it uses, I couldn't even find the packages they are supposed to fix. Thanks again for this!

PR #60, opened 2017-10-30
Fix getting the directory path on Mac OS X.

rbricheno:
With love from PyconUK 2017 sprints!


dw:
Thanks! Good luck with playing with the library. Sorry I'm busy with commercial stuff this week, but if you spam tickets I'll reply in the coming days. :)

PR #62, opened 2017-10-30
[PyCon UK 2017 Sprint] Describe the installation and test process

tai271828:
With love from PyconUK 2017 sprints!

We use test.sh to consolidate test cases in one file so the README
would not be outdated easier rather than the hard-coded test commands
in a code block.

The travis build failure is the same as the original one. It is not a regression.

```
unsocks <module 'six' from '/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/six.pyc'>
F....
FAIL: test_ansible_six_messed_up_path (__main__.BrokenModulesTest)
  File "tests/responder_test.py", line 77, in test_ansible_six_messed_up_path
    self.assertTrue(isinstance(msg.unpickle(), tuple))
AssertionError: False is not true
Ran 5 tests in 0.353s
FAILED (failures=1)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,20 +12,7 @@ install:
 - pip install -r dev_requirements.txt
 
 script:
-- PYTHONPATH=. timeout 05.0 python tests/call_function_test.py
-- PYTHONPATH=. timeout 05.0 python tests/channel_test.py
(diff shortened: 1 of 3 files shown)


dw:
Thanks for this!

PR #63, opened 2017-10-30
Prevent MacOS exception on shutdown of closed socket.

rbricheno:
More from PyconUK 2017 sprint!

The wait_for_port method threw an exception on MacOS. This is because MasOS is more picky about shutting down sockets that are already closed than Linux or Windows. There's a more complete writeup of the issue here (which I copied for my comment):

https://github.com/barryp/py-amqplib/issues/45

We take the option of just hiding this specific error to make tests pass on MacOS.

### DIFF
--- a/tests/testlib.py
+++ b/tests/testlib.py
@@ -83,7 +83,21 @@ def wait_for_port(
                 found = True
                 break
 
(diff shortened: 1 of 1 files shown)


dw:
Thanks again Robert!

PR #65, opened 2017-11-11
Use unittest2 and specific assertX methods

### LINKED ISSUE #64: RFC: Use unittest2 for tests?
David, do you have any preference on adding [unttest2](https://pypi.python.org/pypi/unittest2) as a dev dependency? It would allow using the newer `assert*` methods in tests (e.g. `assertTrue`, `assertIsInstance`) to make tests more descriptive and test failure messages more helpful.

Alternatives:

- Use pytest as the test runner, and use plain assert statements.
- Selectively backport/copy from unittest2 into mitogen's own `testlib.TestCase`

moreati:
This doesn't fix any tests or add any new ones, but it does make the test output more useful for failures.

### Before
```
FAIL: test_ansible_six_messed_up_path (__main__.BrokenModulesTest)
  File "tests/responder_test.py", line 77, in test_ansible_six_messed_up_path
    self.assertTrue(isinstance(msg.unpickle(), tuple))
AssertionError: False is not true
```

### After
```
FAIL: test_ansible_six_messed_up_path (__main__.BrokenModulesTest)
  File "tests/responder_test.py", line 78, in test_ansible_six_messed_up_path
    self.assertIsInstance(msg.unpickle(), tuple)
AssertionError: None is not an instance of <type 'tuple'>
```

Fixes #64

### DIFF
--- a/tests/ssh_test.py
+++ b/tests/ssh_test.py
@@ -53,7 +53,7 @@ class SshTest(testlib.DockerMixin, unittest.TestCase):
         except mitogen.ssh.PasswordError, e:
             pass
 
-        assert e[0] == self.stream_class.password_required_msg
+        self.assertEqual(e[0], self.stream_class.password_required_msg)
 
     def test_password_incorrect(self):
         try:
@@ -65,7 +65,7 @@ class SshTest(testlib.DockerMixin, unittest.TestCase):
         except mitogen.ssh.PasswordError, e:
             pass
 
-        assert e[0] == self.stream_class.password_incorrect_msg
+        self.assertEqual(e[0], self.stream_class.password_incorrect_msg)
 
     def test_password_specified(self):
         context = self.docker_ssh(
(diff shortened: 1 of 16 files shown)


tests/ssh_test.py
```diff
@@ -1,10 +1,9 @@
-
-import unittest
-
 import mitogen
 import mitogen.ssh
 import mitogen.utils
 
+import unittest2 as unittest
```
dw: This looks amazing, but please, no import aliases, it makes code impossible to dive into. If you're using unittest2 please name it explicitly throughout the code, otherwise confusion will abound.

dw:
Hate to be a pain about that kind of stuff, but it gets out of control too easily, so I prefer to avoid it wherever possible, except in some very popular uses like numpy or matplotlib

dw:
Thanks so much for these changes! You're forcing me to keep all the context in my head despite the current barrage of commercial work ;)

moreati:
No problem, and thank you. If I'm becoming a distraction please just say the word and I'll lay off doing PRs for a while.

PR #66, opened 2017-11-11
Replace all calls to file() with open()

moreati:
Although these are synonyms in Python 2.x, when using MyPy to typecheck
code use of file() causes spurious errors.

This commit also serves as one small step to Python 3.x compatibility,
since 3.x removes the file() builtin.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -363,7 +363,7 @@ a large fleet of machines, or to alert the parent of unexpected state changes.
             elif new_size < size:
                 size = 0
 
-            fp = file(path, 'r')
+            fp = open(path, 'r')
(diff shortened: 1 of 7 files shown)


dw:
Thanks for this! Looks like this was supported the whole way through Python 2.4 so we can merge it (until 20 minutes ago I was convinced open() was added in 2.5!). As mentioned previously, 3.x remains a mythical non-target as far as I am concerned, because I've never encountered it being used in real-world infrastructure I actually get paid to work on. Until that changes, I'd prefer to avoid thinking too much about 3.x, but if you know a 3.x related change won't break 2.4, please flag it in the commit :)

PR #80, opened 2018-02-14
Fix/split test_master tests following split of mitogen.master -> mitogen.parent

moreati:


dw:
Just a note on the parent/master front: the current division is a real hatchet job, there are parts of both that probably still belong in the other, as visible in fakessh which needs to implement master in addition to parent.

If the split doesn't make sense to you, don't worry, it really doesn't make sense to anyone :)

PR #82, opened 2018-02-14
Fix TypeError while initiliazing ImporterMixin

moreati:
Since f44356af3275912f7df78fa58e860f107a71e7e5 mitogen.core.Importer()
takes a Router instance.

### DIFF
--- a/tests/importer_test.py
+++ b/tests/importer_test.py
@@ -18,7 +18,7 @@ class ImporterMixin(testlib.RouterMixin):
     def setUp(self):
         super(ImporterMixin, self).setUp()
         self.context = mock.Mock()
-        self.importer = mitogen.core.Importer(self.context, '')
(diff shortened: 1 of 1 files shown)


moreati:
One thing I'm unsure of, `testlib.RouterMixin.router` is still initialized with a `mitogen.master.Router()` instance. Is a `mitogen.parent.Router()` instance more appropriate now?

dw:
It should probably depend on the test being run.

I've developed quite sour grapes for for the interface to Router and Broker and Context vary depending on where you call them, but I don't know any better solution without dumping everything into core.py.

The layering here really sucks, it needs some guiding principle to make things clear. And meanwhile, there is still hacks like that insane "upgrade_router()" that it would _really_ nice to get rid of somehow.

PR #83, opened 2018-02-14
Fix lingering references to mitogen.core.Receiver.get_data()

moreati:


dw:
I __promise__ to start looking at the tests seriously real soon!!

By the way, if you're looking for easy pickings, the new module loader is a joke. It "works" but I can lose count just thinking about bugs I know about :)

PR #93, opened 2018-02-22
Use Python 2.4 compatible module import

### LINKED ISSUE #92: import mitogen raises SyntaxError on Python 2.4.x
```
~/src/mitogen % python -V     
Python 2.4.4
~/src/mitogen % python -c "import mitogen; mitogen.main()"
  File "<string>", line 1, in ?
  File "mitogen/__init__.py", line 113
    from . import utils
         ^
SyntaxError: invalid syntax
```
Patch in progress

moreati:
Python 2.4 does not support explicit relative imports. They were added
at Python 2.5, along with `from __future__ import absolute_import`.

On 2.x this will mean the import is first (implicitly) tried relative,
but on 3.x it will always be tried absolute.

Fixes #92

### DIFF
--- a/mitogen/__init__.py
+++ b/mitogen/__init__.py
@@ -99,7 +99,7 @@ def main(log_level='INFO'):
     def wrapper(func):
         if func.__module__ != '__main__':
             return func
-        from . import utils
+        from mitogen import utils
         utils.log_to_file(level=log_level)
         return utils.run_with_router(func)
     return wrapper


dw:
I wrote a mitogen/docker.py earlier and pointed it at Centos 5, it failed
horribly :( def needs work here. On a related note, any clues about
handling 3.x? My best/least magical is duplicate modules passed through
lib2to3, but this sucks. Another variant is regexing the one feature that
really hurts, which is "except .. as ..". That naturally also sucks, but as
far as hacks go it seems relatively stomachable

moreati:
My plan A for 3.x was convincing you to drop Python < 2.6. Plan B was some form of 2to3 integration, e.g. a setup.py, step or something that ran on Travis and enforced running 2to3 as part of the test suite.

dw:
Your plan may succeed! 2.4 is an ancient stinky target, but I'll consider it a personal failure if it can't be made to work :)

dw:
Holy cowbells, use_2to3=True in setup.py 'just works' (in that we get something 3.x can import but of course bytes/str situation is a mess). Maybe there is a cheap compromise option -- supporting 3.x installation but in the meantime only supporting >2.7x clients with a 3.x-installed library

PR #99, opened 2018-02-26
Test and document semantics of is_blacklisted_module()

moreati:


dw:
I feel so disgracefully guilty having someone else writing tests :/  Promise to get on this before end of March!

moreati:
Don't worry - this is my play time to spend as I choose; I'm learning more by working on the tests than by trying to add features; I think a second pair of eyes results in better tests.

PR #107, opened 2018-03-06
Code golf some bytes from first stage and preamble stage

moreati:
Taken together these commits reduce the first stage from 482 bytes to ~~448~~ 439 bytes, and the preamble from 8946 bytes to 8941 bytes. They eliminate use of `input()` in the first stage, for better Python 3.x compatibility. The use of 'zip' as an alias for the 'zlib' codecs breaks compatibility with Python 3.0 -> 3.3 though.

@dw I'm unsure i precomputing the compressed & uncompressed preamble sizes introduces any race conditions, e.g. could the length of the preamble change between `create_child()`/`get_boot_command()` and `_ec0_received()`?

### DIFF
--- a/mitogen/parent.py
+++ b/mitogen/parent.py
@@ -321,7 +321,10 @@ class Stream(mitogen.core.Stream):
             os.dup2(0,100)
             os.dup2(R,0)
             os.dup2(r,101)
-            for f in R,r,W,w:os.close(f)
+            os.close(R)
+            os.close(r)
+            os.close(W)
+            os.close(w)
             os.environ['ARGV0']=e=sys.executable
             os.execv(e,['mitogen:CONTEXT_NAME'])
         os.write(1,'EC0\n')
--- a/mitogen/parent.py
+++ b/mitogen/parent.py
@@ -317,7 +317,6 @@ class Stream(mitogen.core.Stream):
     # Optimized for minimum byte count after minification & compression.
     @staticmethod
     def _first_stage():
-        import os,sys
         R,W=os.pipe()
         r,w=os.pipe()
         if os.fork():
@@ -352,7 +351,7 @@ class Stream(mitogen.core.Stream):
         # same str (2.x) or an equivalent bytes (3.x).
         return [
             self.python_path, '-c',
-            'import codecs;_=codecs.decode;'
+            'import codecs,os,sys;_=codecs.decode;'
             'exec(_(_("%s".encode(),"base64"),"zip"))' % (encoded,)
         ]
 
--- a/mitogen/parent.py
+++ b/mitogen/parent.py
@@ -331,7 +331,7 @@ class Stream(mitogen.core.Stream):
             os.environ['ARGV0']=e=sys.executable
             os.execv(e,['mitogen:CONTEXT_NAME'])
         os.write(1,'EC0\n')
-        C=_(sys.stdin.read(PREAMBLE_COMPRESSED_LEN), 'zlib')
+        C=_(sys.stdin.read(PREAMBLE_COMPRESSED_LEN), 'zip')
         os.fdopen(W,'w',0).write(C)
         os.fdopen(w,'w',0).write('PREAMBLE_LEN\n'+C)
         os.write(1,'EC1\n')
@@ -353,7 +353,7 @@ class Stream(mitogen.core.Stream):
         return [
             self.python_path, '-c',
             'import codecs;_=codecs.decode;'
-            'exec(_(_("%s".encode(),"base64"),"zlib"))' % (encoded,)
+            'exec(_(_("%s".encode(),"base64"),"zip"))' % (encoded,)
         ]
 
     def get_preamble(self):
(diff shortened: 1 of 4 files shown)


moreati:
I excluded one further change, that I couldn't get to pass the `ssh_test.py` tests

```diff
@@ -331,7 +325,7 @@ class Stream(mitogen.core.Stream):
             os.environ['ARGV0']=e=sys.executable
             os.execv(e,['mitogen:CONTEXT_NAME'])
         os.write(1,'EC0\n')
-        C=_(sys.stdin.read(PREAMBLE_COMPRESSED_LEN), 'zip')
+        C=_(os.read(0, PREAMBLE_COMPRESSED_LEN), 'zip')
         os.fdopen(W,'w',0).write(C)
         os.fdopen(w,'w',0).write('PREAMBLE_LEN\n'+C)
         os.write(1,'EC1\n')
```

If it could be made to work then `os.read(0, ...)` 
1. eliminates 1 of the 2 remaining uses of the sys module in the first stage
1. returns a `bytes` object we can pass to `codecs.decode(..., 'zip') in Python 3.x

`os.fdopen(0).read(...)` does pass the tests. It has benefit 1, but not benefit 2. Using it doesn't alter the first stage size.

mitogen/parent.py
```diff
@@ -328,16 +331,19 @@ def _first_stage():
...
+        os.fdopen(w,'w',0).write('PREAMBLE_LEN\n'+C)
         os.write(1,'EC1\n')
 
     def get_boot_command(self):
         source = inspect.getsource(self._first_stage)
         source = textwrap.dedent('\n'.join(source.strip().split('\n')[2:]))
         source = source.replace('    ', '\t')
         source = source.replace('CONTEXT_NAME', self.remote_name)
+        preamble_compressed = self.get_preamble()
+        source = source.replace('PREAMBLE_COMPRESSED_LEN', str(len(preamble_compressed)))
```
dw: This is an awesome simplification!

dw:
Note os.read(n) empties up to n bytes from whatever is available in the kernel at the time, fileobj.read(n) repeatedly calls os.read() until n bytes have been read, blocking repeatedly.

We always need to guard against the possibility either the buffers can't hold the full core.py zlibbed source, or timing causes only half of it to be delivered (don't forget there are all kinds of weird timers in TCP like in Nagle's algorithm, that might cause a busy box to transmit at odd times)

Doing one final pass of review and pomp reduction on the post, so will merge this later :)

moreati:
Thank you. An idea: if connection specific data was moved from the exec'd
portion to the surrounding code then the cpu cost of minimising/compressing
it could be incurred just once rather than per connection. At the expense
of a few bytes longer first stage

moreati:
I'm done tinkering with this. I considered one more tweak, but I think it would interact badly with the [Mac OS workaround](https://github.com/dw/mitogen/commit/2b5e40b7086e33ed8fb4663babb311231907e434)

```diff
@@ -312,8 +306,8 @@ class Stream(mitogen.core.Stream):
     # Minimised, gzipped, base64'd and passed to 'python -c'. It forks, dups
     # file descriptor 0 as 100, creates a pipe, then execs a new interpreter
     # with a custom argv.
-    # 'CONTEXT_NAME', 'PREAMBLE_COMPRESSED_LEN', and 'PREAMBLE_LEN' are
-    # substituted with their respective values.
+    # 'CONTEXT_NAME', 'PREAMBLE_COMPRESSED_LEN', 'PREAMBLE_LEN', and 'PYTHON'
+    # are substituted with their respective values.
     # Optimized for minimum byte count after minification & compression.
     @staticmethod
     def _first_stage():
@@ -327,8 +321,8 @@ class Stream(mitogen.core.Stream):
             os.close(r)
             os.close(W)
             os.close(w)
-            os.environ['ARGV0']=sys.executable
-            os.execl(sys.executable,'mitogen:CONTEXT_NAME')
+            os.environ['ARGV0']='PYTHON'
+            os.execlp('PYTHON','mitogen:CONTEXT_NAME')
         os.write(1,'EC0\n')
         C=_(os.fdopen(0,'rb').read(PREAMBLE_COMPRESSED_LEN),'zip')
         os.fdopen(W,'w',0).write(C)
@@ -343,6 +337,7 @@ class Stream(mitogen.core.Stream):
         preamble_compressed = self.get_preamble()
         source = source.replace('PREAMBLE_COMPRESSED_LEN', str(len(preamble_compressed)))
         source = source.replace('PREAMBLE_LEN', str(len(zlib.decompress(preamble_compressed))))
+        source = source.replace('PYTHON', self.python_path)
         encoded = zlib.compress(source, 9).encode('base64').replace('\n', '')
         # We can't use bytes.decode() in 3.x since it was restricted to always
         # return unicode, so codecs.decode() is used instead. In 3.x
@@ -351,7 +346,7 @@ class Stream(mitogen.core.Stream):
         # same str (2.x) or an equivalent bytes (3.x).
         return [
             self.python_path, '-c',
-            'import codecs,os,sys;_=codecs.decode;'
+            'import codecs,os;_=codecs.decode;'
             'exec(_(_("%s".encode(),"base64"),"zip"))' % (encoded,)
         ]
```

dw:
Will try and get this one merged tomorrow finally. Sorry for the delay

moreati:
I'm in no rush. It's bound to break _something_, and the test suite is struggling to be useful while #121 is open.

PR #133, opened 2018-03-12
Fixed documentation error for async operation

boxed:


dw:
Thanks for this! There are actually a couple more breakages on index.rst from recent changes. I'm currently battling fallout from a 'fix' to the Ansible stuff that's totally broken single-file programs. Will merge this once that is fixed, and I can test the change locally ;)

PR #142, opened 2018-03-14
tests: Add Tox config for Python 2.6 and 2.7

moreati:
I could not get Python 2.5 or earlier to work. Too many packages (critically docker) don't support it.

[pyenv](https://github.com/pyenv/pyenv) or it's plugin [python-build](https://github.com/pyenv/pyenv/tree/master/plugins/python-build) are the best ways I know to install different Python versions.


dw:
Thanks Alex!

PR #167, opened 2018-03-25
Add testimonial from Clojars

danielcompton:


dw:
Thanks a ton! Going to take your tweet version instead, don't like having my name quite so visible on the page :)

Issue #102, opened 2018-02-27
ansible: tests for every playbook keyword


dw:
Full list is here: http://docs.ansible.com/ansible/latest/playbooks_keywords.html

dw:
Per #175, every test Mitogen has for functionality like sudo(password=...) must also be duplicated in the form of an Ansible playbook

PR #175, opened 2018-03-30
Use become_pass for sudo password

wezm:
With the recent fixes for the [tty issue](https://github.com/dw/mitogen/issues/171) I was still unable to successfully apply a playbook to my test machines. These particular machines are accessed via a normal user via an ssh authorized key. sudo requires a password and this was supplied via the `-K` option when invoking `ansible-playbook`. The error I was getting was:

```
CallError: mitogen.core.CallError: mitogen.sudo.PasswordError: sudo password is required
```

After much digging I found https://github.com/ansible/ansible/blob/fd96bcd22faf3656f0f830203841bc07d0e22937/lib/ansible/playbook/play_context.py#L492 and noticed that mitogen was using `password` in the play context instead of `become_pass`. The change in this PR allows me to successfully apply the playbook to the machines (but no doubt needs some finessing and tests before being ready to merge).

### DIFF
--- a/ansible_mitogen/connection.py
+++ b/ansible_mitogen/connection.py
@@ -198,7 +198,7 @@ class Connection(ansible.plugins.connection.ConnectionBase):
             cast({
                 'method': 'sudo',
                 'username': self._play_context.become_user,
-                'password': self._play_context.password,
+                'password': self._play_context.become_pass,
                 'python_path': python_path or self.python_path,
                 'sudo_path': self.sudo_path,
                 'connect_timeout': self._play_context.timeout,


dw:
You're a hero! Ansible has some higher level code that does some of this variable merging on our behalf, so it's not entirely clear what should be used where. :)  This will have real tests added during the work in #102. Thanks a ton for doing the legwork here, and sorry for the inconvenience.

Please also feel free to share timing information and spam as many one-liner bug reports/emails as you like -- naturally I'm particularly interested in mediocre results coming from backers. Aside from multi-target runs (which are sluggish but not slow), genuinely slow playbooks (big apt-get etc.), and Ansible 2.5.0 (perf regression in mainline), I'm interested in anything that looks like it shouldn't be slow but currently is.

Issue #160, opened 2018-03-22
RFC: docstrings


dw:
In favour of removal:

* Can finally eliminate minimize_source()
* Easy to minimize wire footprint, even for non core.py modules that don't get minimize_source() treatment just now

Cons of removal:

* Code looks ugly and unprofessional
* Pain in the ass for new contributors to understand any of the code
* Pain in the ass to keep Sphinx equivalents in sync, and no good place in the documentation for all kinds of random internal-only methods.

First alternative idea:

* Just accept source manipulation is a thing, because it lets us have our cake (wire footprint) and eat it (readable source).
* Have ModuleResponder scan for ``MITOGEN_MINIMIZE_SAFE = True`` or similar, and respond to that by doing horrible things only to the source of modules that have explicitly elected for such treatment. The reason it needs to be opt-in is because, even if it weren't for the code style-sensitive regexes, there is a /lot/ of code that introspects ``__doc__`` to implement behaviours of some sort or other (simple example: doctest) 
* Sleep better at night perhaps by building something (hopefully speedy) around the ``parser`` module, to avoid all those insane regexes.

@moreati? I've a feeling I know which way you'll swing already :)

moreati:
> @moreati? I've a feeling I know which way you'll swing already :)

Cake please. +1 to keep docstrings, and either accept the overhead or use a stripper/minifier that is based on parsing/tokenising.

For reference these are the numbers with and without `minimize_source`

```
:~/src/mitogen p-p-p-pick-us-a-pickle(+140/-45)+*± python preamble_size2.py 
SSH command size: 417
Preamble size: 9297 (9.08KiB)
     Module        Original       Minimized       Compressed   
                Bytes  KiB  %    Bytes  KiB  %    Bytes  KiB  %   
mitogen.master  24070 23.5  100  15332 15.0 63.7   4743  4.6 19.7
mitogen.parent  18835 18.4  100  12834 12.5 68.1   4196  4.1 22.3
mitogen.ssh      4986  4.9  100   2788  2.7 55.9   1028  1.0 20.6
mitogen.sudo     6224  6.1  100   3054  3.0 49.1   1093  1.1 17.6
mitogen.fakessh 11588 11.3  100   7591  7.4 65.5   2518  2.5 21.7

:~/src/mitogen p-p-p-pick-us-a-pickle(+140/-45)+*± python preamble_size2.py 
SSH command size: 417
Preamble size: 9297 (9.08KiB)
[...]

PR #178, opened 2018-03-31
Reimplement minimize_source() using tokenize

moreati:
### Benefits:

- Much ~~more correct~~ less broken than re.sub()
- Handles trailing whitespace
- Recognises doc-strings regardless of quoting style

### Limitations:

- Still not entirely correct
  - Creates a syntax error when function/class/loop/... body is _only_ a docstring
- Slower by 50x - 18 ms vs 0.34 ms for re.sub() to minimize `mitogen.master`
  - Not much scope for improving this, tokenize is 100% pure Python
- Complex state machine, harder to understand
- Higher line count in parent.py
- Untested with Mitogen parent on Python 2.x and child on Python 2.x+y

### No change

- Only requires Python stdlib modules
- Wire size of modules (except `mitogen.parent`) is roughly the same

### Before

```
SSH command size: 421
Preamble size: 9771 (9.54KiB)
                  Original          Minimized           Compressed     
mitogen.master  23234 22.7KiB  14608 14.3KiB 62.9%   4487  4.4KiB 19.3%
mitogen.parent  27293 26.7KiB  18992 18.5KiB 69.6%   5749  5.6KiB 21.1%
mitogen.ssh      5042  4.9KiB   2832  2.8KiB 56.2%   1035  1.0KiB 20.5%
mitogen.sudo     6183  6.0KiB   3053  3.0KiB 49.4%   1093  1.1KiB 17.7%
mitogen.fakessh 11545 11.3KiB   7586  7.4KiB 65.7%   2508  2.4KiB 21.7%
```

### After

```
SSH command size: 421
Preamble size: 9625 (9.40KiB)
                  Original          Minimized           Compressed     
mitogen.master  23234 22.7KiB  14589 14.2KiB 62.8%   4376  4.3KiB 18.8%
mitogen.parent  29956 29.3KiB  21151 20.7KiB 70.6%   6125  6.0KiB 20.4%
mitogen.ssh      5042  4.9KiB   2863  2.8KiB 56.8%   1023  1.0KiB 20.3%
mitogen.sudo     6183  6.0KiB   3089  3.0KiB 50.0%   1089  1.1KiB 17.6%
mitogen.fakessh 11545 11.3KiB   7655  7.5KiB 66.3%   2488  2.4KiB 21.6%
```

Based on idea in https://stackoverflow.com/a/1769794/293340

Fixes issue #160


dw:
Totes exhausted today, but already this is looking awesome :) Could be a good opportunity to move the minimize_source stuff out of parent into its own file, if that's easy.

The source minimization stuff ideally only needs to live on the master -- once the minimized source starts being passed into child Importer caches, there is no point in re-minimizing of course

I can't remember why I kept minimize_source() in there.. I suspect something about moving it was annoying :)  Already thinking of splitting more stuff out e.g. the dep scanner bits in master.py.

Ideally we can arrange every module 'logically' except for core.py, and the importer will just DTRT (and not spend extra roundtrips), and if it doesn't then that's a bug

dw:
Haven't forgotten! Just need to check it out and kick the tires a bit before merging :)

PR #180, opened 2018-04-01
Switch to using unit2 test runner

moreati:
This means test files are imported as modules, not run as scripts. THey can still be run individually if so desired. Test coverage is measured, and an html report generated in htmlcov/. Test cases are automatically discovered, so they need not be listed in two places. An overall passed/failed/skipped summary is printed, rather than for each file.

Arguments passed to ./test are passed on to unit2. For instance

    ./test -v

will print each test name as it is run.


dw:
Insta-merge!   One tweak I'll put on top is moving test to ./run_tests or something.. test vs tests makes bash tab complete super annoying (and it's been broken since test.sh was added)

PR #248, opened 2018-05-14
fix host_keys= exception message

badeadan:


dw:
re: IRC, sorry, I haven't had a chance to look at fakessh yet, sorry. Stuck on some really nasty Ansible bits.

PR #269, opened 2018-06-18
Fix Docker image construction and Tox test invocation

moreati:


dw:
Thanks! Sorry for delay

dw:
Been carrying around that for loop in a git stash for weeks :S

PR #303, opened 2018-07-13
Support for "doas" become_method

napkindrawing:
Howdy!

I've hacked up a first-pass at basic support for the `doas` `become_method`, which is working for me on FreeBSD.

I haven't added any unit tests yet, as I'm not familiar with how the tests are laid out here.

Cheers,
Mike

### DIFF
--- a/ansible_mitogen/connection.py
+++ b/ansible_mitogen/connection.py
@@ -173,6 +173,20 @@ def _connect_sudo(spec):
     }
 
 
+def _connect_doas(spec):
+    return {
+        'method': 'doas',
+        'enable_lru': True,
+        'kwargs': {
+            'username': spec['become_user'],
+            'password': wrap_or_none(mitogen.core.Secret, spec['become_pass']),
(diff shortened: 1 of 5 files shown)


dw:
Champion! The structure of the connection plug-ins will drastically change for asynchronous connect when that arrives, so right now what you have is perfect -- a little cutpaste it's all I've been doing too :P

Eventually these will be converted to a small set of state machine classes, and the most important thing to capture are the expected strings and general behaviour.

I need to write better stub test implementations for a bunch of stuff, including su, so I will add doas to the list.

Also the "mitogen" Travis jobs are failing right now due to the child process reaping issue -- don't worry about those either.

napkindrawing:
Glorious!

Thanks so much again for all your work with mitogen. It's such an exciting piece of software on it's own, and I still can't believe what a huge impact it will have on ansible  🎉

dw:
Thanks for the PR! It's on master now, I hope to cut another release in the coming week once a few more tickets are closed

PR #318, opened 2018-07-24
Fixed ProcessLookupError issue when _reap_child is called.

tbtmuse:


dw:
Hi there. Thanks for the PR :)

This test has been left broken and enabled in order to understand *why* ESRCH is returned sometimes. It seems that something else in the process might be calling `waitpid(0)`, so swallowing the exception here is likely hiding a bug elsewhere. The problem is that I have been unable to reproduce it except on Travis CI.

[Looking at an example log](https://api.travis-ci.org/v3/job/407403656/log.txt), the only place I can think of is `class ProcessMonitor` in `parent.py` which is activated by way of `fakessh_test.py`, which is called before `parent_test` in that log. But reading the ProcessMonitor code, it's not obvious how calling `waitpid()` for only explicitly registered PIDs could trigger it. Regardless that is the first place I would go digging.

Feel free to have a shot at figuring this one out! If you can reproduce the failure on your local machine, then running the tests under [strace](https://linux.die.net/man/1/strace) might be sufficient to at least spot the stray `waitpid()` call, if not also help identify where it's coming from.

dw:
I managed to tickle it running a command line like:

```
while : ; do rm -f /tmp/trace* ; MITOGEN_LOG_LEVEL=debug strace -o /tmp/trace -ff bash ./run_tests -f || break ; done
```

The last output from the tests are:

```
16:00:33.181769 D mitogen: tty_create_child() child 50980 fd 53, parent 50702, cmd: bash -c "exec 2>/tmp/tmpJ3fXXV; echo hi > /dev/tty"
E
ERROR: test_dev_tty_open_succeeds (parent_test.TtyCreateChildTest)
  File "/home/dmw/src/mitogen/tests/parent_test.py", line 130, in test_dev_tty_open_succeeds
    waited_pid, status = os.waitpid(pid, 0)
OSError: [Errno 10] No child processes

Ran 122 tests in 11.412s

FAILED (errors=1, skipped=2)
```

Grepping for PID 50980 in `/tmp/trace.*` reveals some confusing things:

```
trace.50704
20:--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=50980, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---

trace.50702
36819:clone(child_stack=0, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7f22b5f1a9d0) = 50980
36825:wait4(50980, [{WIFEXITED(s) && WEXITSTATUS(s) == 0}], WNOHANG, NULL) = 50980
36831:wait4(50980, 0x7fff112bd3ac, 0, NULL)   = -1 ECHILD (No child processes)

trace.50980
7:setsid()                                = 50980
1117:getpid()                                = 50980
1161:getpgrp()                               = 50980
```

PID 50704 is receiving SIGCHLD but it never calls waitpid. From the remainder of the strace output, it looks like this is an idle broker thread, but I'm not sure. In Python on Linux, signals are caught and pushed on to a queue to be handled by the main thread of a process, which in this case is pid 50702 (found by grepping for 50704 in the trace output).

50702 looks like the main thread of the failing test, and indeed we can see the stray waitpid!

```
write(2, "16:00:33.181769 D mitogen: tty_c"..., 136) = 136
wait4(50980, [{WIFEXITED(s) && WEXITSTATUS(s) == 0}], WNOHANG, NULL) = 50980
epoll_create(32)                        = 57
epoll_ctl(57, EPOLL_CTL_ADD, 53, {EPOLLIN, {u32=53, u64=12937519772861464629}}) = 0
epoll_wait(57, [{EPOLLIN|EPOLLHUP, {u32=53, u64=12937519772861464629}}], 32, 4999) = 1
read(53, "hi\n", 4096)                  = 3
close(57)                               = 0
wait4(50980, 0x7fff112bd3ac, 0, NULL)   = -1 ECHILD (No child processes)
```

Notice the double wait4(). But I still have no idea how this call is being generated. `_reap_child()` calls waitpid with WNOHANG, but so does that signal handler in ProcessMonitor.

I wonder if we try simply deleting or commenting out ProcessMonitor, if the problem will go away? In that case, if it's just ProcessMonitor that is broken, disabling the fakessh tests for now might be the real fix -- that code requires a big redesign and it's not in scope for 0.2.x anyway :)

tbtmuse:
Hi, amazing tool by the way, you have done some excellent work here, forgot to mention that!!!!!

Now in response to your previous post. I have tried commenting out ProcessMonitor, as it indeed does look like the culprit, but to my surprise that has had no effect at all. I still get the same error.

Further investigation is required....

dw:
I think I found it! It's the `subprocess.Popen.__del__` destructor. If the child exits before `__del__` is called, then its attempt to waitpid() will succeed, stealing the process exit status.

Ways to fix this:

- Make children always outlive `__del__`.. not possible
- Rip out subprocess.. it's there primarily to handle the annoying pickling/unpickling of errors raised in the forked child. That code is a pain to replace, so keeping it seems more practical.
- Move to using subprocess.Popen instances to manage subprocesses everywhere.. this is probably the ideal route, but it requires interface changes throughout the library, which aren't sensible in a stable series
- Mudge `Popen._child_created` attribute from within `tty_create_child()` and friends.. it's a hack, but it fixes exactly the issue.

I'm not quite ready to fully depend on Popen everywhere, based at least on the fact it has surprising behaviour like this. Quite happy to accept the `_child_created` hack above if you'd like to try it out :)

tbtmuse:
Brilliant let me give it a shot, will get back to you.

tbtmuse:
Hi sorry for the silence, a little tied up with work. Could you kindly provide a little more guidance as to how I can do this: "Mudge Popen._child_created attribute from within tty_create_child() and friends.. it's a hack, but it fixes exactly the issue." looking into this now.

dw:
I was thinking of literally just inserting 'proc._child_created = False' somewhere in tty_create_child(). It's a hack, but it avoids having to change the Stream internals during a stable release.

tbtmuse:
No luck, adding 'proc._child_created = False' results in no change, Still getting "No such process" randomly.

I tried refactoring the code a bit to allow Popen.poll(), which sets the returncode and returns None if the process is still in the process table. That seemed like a viable solution but that had no effect.

Popen.terminate() essentially does exactly what the code is doing now, except it checks the returncode of the process before sending os.kill(), however the result is the same. 

For now the only way I can avoid the error reliably is catching the "No such process" which does look like it is solely caused by "os.kill(self.pid, signal.SIGTERM)"

Calling waitpid with os.WNOHANG returns (0, 0), more often than not, which means "no child process status is immediately available" what this means exactly isn't clear.

tbtmuse:
I think I have managed to resolve my issue by registering a signal handler.

I might be going on a complete tangent here but please let me know what you think.

DavidVentura:
I believe I am experiencing this issue..

```
ERROR! [mux  28528] 09:47:10.844462 E mitogen.ctx.ssh.host: mitogen: mitogen.fork.Stream('fork.13988') crashed
  File "<stdin>", line 2959, in _call
  File "<stdin>", line 1629, in on_receive
  File "master:/home/david/git/pcon/ansible/mitogen-0.2.6/mitogen/parent.py", line 1290, in on_disconnect
    self._reap_child()
  File "master:/home/david/git/pcon/ansible/mitogen-0.2.6/mitogen/parent.py", line 1280, in _reap_child
    os.kill(self.pid, signal.SIGTERM)
ProcessLookupError: [Errno 3] No such process
ERROR! [mux  28528] 09:47:10.854249 E mitogen.ctx.ssh.host2: mitogen: mitogen.fork.Stream('fork.17372') crashed
  File "<stdin>", line 2959, in _call
  File "<stdin>", line 1629, in on_receive
  File "master:/home/david/git/pcon/ansible/mitogen-0.2.6/mitogen/parent.py", line 1290, in on_disconnect
    self._reap_child()
  File "master:/home/david/git/pcon/ansible/mitogen-0.2.6/mitogen/parent.py", line 1280, in _reap_child
    os.kill(self.pid, signal.SIGTERM)
ProcessLookupError: [Errno 3] No such process
```

I am on 0.2.6

PR #350, opened 2018-09-05
Python 3 support for classmethod call targets

### LINKED ISSUE #313: cannot pass class-bound method to context call
As described by the [documentation](https://mitogen.readthedocs.io/en/stable/api.html#mitogen.parent.Context.call_async), I expect to be able to specify as a call target of `Context.call_async` – `fn` – either a:

* module-level function
* static method
* class method

However, at least in naively passing the Python object reference of either version of a method, mitogen is either unable to find or unable to process the call target.

For example, using `staticmethod` –

    #!/usr/bin/env python
    import os
    import pathlib
    import socket
    
    import mitogen.utils
    
    class MyClass:
    
        @staticmethod
        def my_first_function():
            print('hostname:', socket.gethostname())
            print('remote PID:', os.getpid())
            print('file:', pathlib.Path(__file__))
            return 123
    
        @classmethod
        def main(cls, router):
            local = router.local()
            result = local.call(cls.my_first_function)
            print(result, socket.gethostname(), os.getpid())
    
    if __name__ == '__main__':
        mitogen.utils.log_to_file()
        mitogen.utils.run_with_router(MyClass.main)

– results in the following error –

      File "./fake_manage_method.py", line 27, in <module>
        mitogen.utils.run_with_router(MyClass.main)
      File "/home/USER/dev/repo/opt/mitogen/mitogen/utils.py", line 100, in run_with_router
        return func(router, *args, **kwargs)
      File "./fake_manage_method.py", line 21, in main
        result = local.call(cls.my_first_function)
      File "/home/USER/dev/repo/opt/mitogen/mitogen/parent.py", line 1016, in call
        return receiver.get().unpickle(throw_dead=False)
      File "/home/USER/dev/repo/opt/mitogen/mitogen/core.py", line 487, in unpickle
        raise obj
    mitogen.core.CallError: builtins.AttributeError: module '__main__' has no attribute 'my_first_function'
      File "<stdin>", line 2049, in _dispatch_calls
      File "<stdin>", line 2036, in _dispatch_one

Or, replacing the `staticmethod` with `classmethod`:

      File "./fake_manage_method.py", line 27, in <module>
        mitogen.utils.run_with_router(MyClass.main)
      File "/home/USER/dev/repo/opt/mitogen/mitogen/utils.py", line 100, in run_with_router
        return func(router, *args, **kwargs)
      File "./fake_manage_method.py", line 21, in main
        result = local.call(cls.my_first_function)
      File "/home/USER/dev/repo/opt/mitogen/mitogen/parent.py", line 1015, in call
        receiver = self.call_async(fn, *args, **kwargs)
      File "/home/USER/dev/repo/opt/mitogen/mitogen/parent.py", line 1012, in call_async
        return self.send_async(make_call_msg(fn, *args, **kwargs))
      File "/home/USER/dev/repo/opt/mitogen/mitogen/parent.py", line 401, in make_call_msg
        isinstance(fn.im_self, (type, types.ClassType)):
    AttributeError: 'function' object has no attribute 'im_self'

In the case of the `staticmethod`, (unpickling), this may be worked around, for example –

    def my_first_function():
        print('hostname:', socket.gethostname())
        print('remote PID:', os.getpid())
        print('file:', pathlib.Path(__file__))
        return 123
    
    class MyClass:
    
        my_first_function = staticmethod(my_first_function)
    
        @classmethod
        def main(cls, router):
            local = router.local()
            result = local.call(cls.my_first_function)
            print(result, socket.gethostname(), os.getpid())    

– such that mitogen can now find the call target at the module level; (however, if mitogen intends to support static methods, this is less than ideal).

jesteria:
There were two problems with detection and handling of class methods as call targets in Python 3:

* Methods no longer define `im_self` -- this is now only `__self__`
* The `types` module no longer defines a `ClassType`

The universally-compatible (v2.6+) solution was to switch to using the `inspect` module -- whose interface has been stable -- and to checking the method attribute `__self__`.

(It doesn't hurt that `inspect` checks are more brief and we now no longer need the `types` module here.)

Resolves #313 

---

Note: Though this appears to do the trick --

1. Of course I don't know that I've added the test in the optimal place, _etc._ -- so let me know.
2. ~I'm not clear how the library is being tested against Python 3 -- I don't see it listed in `tox.ini` nor anything like 2to3.~

### DIFF
--- a/mitogen/parent.py
+++ b/mitogen/parent.py
@@ -47,7 +47,6 @@ import termios
 import textwrap
 import threading
 import time
-import types
 import zlib
 
 # Absolute imports for <2.5.
@@ -490,9 +489,8 @@ def upgrade_router(econtext):
 
 
 def make_call_msg(fn, *args, **kwargs):
-    if isinstance(fn, types.MethodType) and \
-       isinstance(fn.im_self, (type, types.ClassType)):
-        klass = mitogen.core.to_text(fn.im_self.__name__)
+    if inspect.ismethod(fn) and inspect.isclass(fn.__self__):
+        klass = mitogen.core.to_text(fn.__self__.__name__)
     else:
         klass = None
 
--- a/tests/call_function_test.py
+++ b/tests/call_function_test.py
@@ -36,13 +36,29 @@ def func_accepts_returns_sender(sender):
     return sender
 
 
+class TargetClass:
+
+    offset = 100
+
+    @classmethod
+    def add_numbers_with_offset(cls, x, y):
+        return cls.offset + x + y
+
+
 class CallFunctionTest(testlib.RouterMixin, testlib.TestCase):
+
     def setUp(self):
         super(CallFunctionTest, self).setUp()
         self.local = self.router.fork()
 
     def test_succeeds(self):
         self.assertEqual(3, self.local.call(function_that_adds_numbers, 1, 2))
+
+    def test_succeeds_class_method(self):
+        self.assertEqual(
+            self.local.call(TargetClass.add_numbers_with_offset, 1, 2),
+            103,
+        )
 
     def test_crashes(self):
         exc = self.assertRaises(mitogen.core.CallError,


jesteria:
Ah hah -- I see, though Python 3 isn't listed in `tox.ini`, it's included in the Travis build :+1:

dw:
Thanks a million for this Jesse! Ignore the broken CI, I've mostly fixed it on my branch but it's not merged down yet due to other breakage.

PR #372, opened 2018-09-17
Use `lxc exec --mode=noninteractive` which is more widely compatible

### LINKED ISSUE #371: mitogen_lxd plugin uses incompatible "lxc --force-noninteractive" flag
* mitogen version: master 49399dfa (includes lxd fix #339)
* ansible 2.5.9, python 2.7.12
* host OS: Ubuntu 16.04.5, which has the old stable version of lxd, 2.0.11

mitogen tries to run `lxc exec --force-noninteractive`, but that flag is not recognised by this version of lxd:

```
ansible-playbook test.yml

PLAY [gold-master] ******************************************************************************************************************

TASK [Gathering Facts] **************************************************************************************************************
fatal: [gold-master]: UNREACHABLE! => {"changed": false, "msg": "EOF on stream; last 300 bytes received: u'error: flag provided but not defined: --force-noninteractive\\n'", "unreachable": true}
	to retry, use: --limit @/home/ubuntu/ansible/misc/test.retry

PLAY RECAP **************************************************************************************************************************
gold-master                : ok=0    changed=0    unreachable=1    failed=0
```

The solution is to use the backwards-compatible `--mode=noninteractive` instead.  Changing that line in [mitogen/lxd.py](https://github.com/dw/mitogen/blob/master/mitogen/lxd.py#L66) fixes the problem.

This flag has evolved over time. There's no manpage on ubuntu 16.04, but here is the online help:

```
$ lxc exec --help
Usage: lxc exec [<remote>:]<container> [--mode=auto|interactive|non-interactive] [--env KEY=VALUE...] [--] <command line>

Execute commands in containers.

Mode defaults to non-interactive, interactive mode is selected if both stdin AND stdout are terminals (stderr is ignored).

Options:
    --debug  (= false)
        Enable debug mode
    --env  (= [])
        Environment variable to set (e.g. HOME=/home/foo)
    --force-local  (= false)
        Force using the local unix socket
    --mode (= "auto")
        Override the terminal mode (auto, interactive or non-interactive)
    --no-alias  (= false)
        Ignore aliases when determining what command to run
    --verbose  (= false)
        Enable verbose mode
```

* [17.10](http://manpages.ubuntu.com/manpages/artful/man1/lxc.exec.1.html) gained `-T` / `-t` flags
* [18.04](http://manpages.ubuntu.com/manpages/bionic/man1/lxc.exec.1.html) gained long aliases `--force-noninteractive` / `--force-interactive`

I checked in the [lxd source code](https://github.com/lxc/lxd/blob/master/lxc/exec.go#L164) that `--mode=noninteractive` and `--force-noninteractive` do the same thing.

There is also a [mutual exclusion test](https://github.com/lxc/lxd/blob/master/lxc/exec.go#L125)

candlerb:
Closes #371

### DIFF
--- a/mitogen/lxd.py
+++ b/mitogen/lxd.py
@@ -63,7 +63,7 @@ class Stream(mitogen.parent.Stream):
         bits = [
             self.lxc_path,
             'exec',
-            '--force-noninteractive',
+            '--mode=noninteractive',
             self.container,
             '--',
         ]
--- a/tests/lxd_test.py
+++ b/tests/lxd_test.py
@@ -18,7 +18,7 @@ class FakeLxcTest(testlib.RouterMixin, unittest2.TestCase):
         argv = eval(context.call(os.getenv, 'ORIGINAL_ARGV'))
         self.assertEquals(argv[0], lxc_path)
         self.assertEquals(argv[1], 'exec')
-        self.assertEquals(argv[2], '--force-noninteractive')
+        self.assertEquals(argv[2], '--mode=noninteractive')
         self.assertEquals(argv[3], 'container_name')
 
 


dw:
Don't worry about CI, it's broken for a variety of reasons, but most importantly, it'll be 100% green before 0.2.3 is release. Thanks for contributing this!

PR #512, opened 2019-02-01
Add mitogen_host variable as optional target

cloudnull:
The option `mitogen_host` has been added allowing users to define it as
a given target which is useful in inventories where the `ansible_host`
is already defined and may be an IP address.

> This option was added to cater for the situation where containers are
  in a givne ansible inventory but are also assigned an IP address. The
  `mitogen_host` option can now be used to point mitogen at a given
  container instead of assuming `ansible_host` is the name of the
  container. This was developed with OpenStack-Ansible in mind and has
  been validated using LXC and nspawn containers which have multiple
  IP addresses.

Documentation has been updated to show how `mitogen_host` is used and
will fall back to the default: `ansible_host` and `inventory_hostname`.

osa-containers test inventory has been updated to reflect the new
option and how it is used.

Signed-off-by: Kevin Carter <kevin.carter@rackspace.com>

done.

  docs/api.rst.
updated docs/ansible.rst

Updated tests/ansible/hosts/osa-containers though no additional tests have been added

n/a

### DIFF
--- a/ansible_mitogen/transport_config.py
+++ b/ansible_mitogen/transport_config.py
@@ -299,7 +299,10 @@ class PlayContextSpec(Spec):
         return self._inventory_name
 
     def remote_addr(self):
-        return self._play_context.remote_addr
+        return (
+            self._connection.get_task_var('mitogen_host') or
+            self._play_context.remote_addr
+        )
 
     def remote_user(self):
         return self._play_context.remote_user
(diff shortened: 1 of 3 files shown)


cloudnull:
abandoning in favor of - https://github.com/dw/mitogen/commit/01840fd0b0c4f152c9a9ebc72d14b7e5fd9de701

PR #515, opened 2019-02-01
Add python interpreter check and transfer

cloudnull:
This change checks the value of the provided python interpreter to see
if it contains an operating system separator. If it does the default
behaviour to parse the given string is executed and retunred. If no
separator is found the string will now be returned as-is and assumed to
be a templated option which ansible internals will handle natively.

This also adds the `mitogen_python_interpreter` which will default to
the `ansible_python_interpreter` when unset. This will allow deployers
to set python interpreters for mitogen as a constant which is important
as the python interpreter may also be set at the task level.

Signed-off-by: Kevin Carter <kevin.carter@rackspace.com>

  docs/api.rst.


dw:
Just had another quick look -- if we do use Templar here, we also need a method to recover it from a 'meta:' statement. They're hard-wired in the guts of `strategy/__init__.py` and as you can see from `_reset_find_task_vars`, getting access to its state is not pretty.

IIRC it also creates a templar, so if we really need to go this route, a similar hack could be used to retrieve it

cloudnull:
I've not been able to figure out a way to render things like `ansible_python_interpreter` without calling templar directly. With the `os.sep` check in the python interpreter function I'm able to see the full variable in the runtime however it never expands, even when querying `task_vars`. That said, I'll keep digging and see if there's a way to use an already instantiated templar to accomplish what is needed.

dw:
I'm happy with the idea that, at least for now, we may need templar. But its use should be moved behind the only variables that definitely require it -- those dicts are recreated for every item, and they're already quite capable of getting under 1ms with the right config ;)

I'd suggest just passing the templar you have through to the transport_config.*Spec constructor, and writing the explicit "if templar" in the individual method for the variables that want it. I know this is messy and lacking in generality, but randomly expanding variables in every context is a potent, ruinous and highly addictive drug  :)  Many of Ansible's woes can be traced to it, as can many old Mitogen bugs!

Don't worry about the meta stuff, I can open a separate ticket for it. It needs a little more work for related reasons, so I'm happy to look at that bit

cloudnull:
Excellent, i'll try and kick an update out shortly.

s1113950:
Some similar things to what's in this PR already landed in #658 ; feel free to update your PR with `master` and we can merge what's left! :)

PR #566, opened 2019-03-15
docs: fixed message routing example description

lumatijev:
Message routing example description does not match given diagram. Or maybe I missed something? :blush:


dw:
Thanks a ton for these! I'm very good at missing 'subtle' details like this :)

lumatijev:
Glad I could help. Keep up good work! :smiley:

PR #595, opened 2019-06-09
Add buildah transport

jordemort:
This adds a transport for [Buildah](https://github.com/containers/buildah/) - with this and a little bit of hacking, I am able to accelerate [ansible-bender](https://github.com/ansible-community/ansible-bender/) with Mitogen. This brings my container build time down from 12 minutes to 5 minutes.

TODO:
  - [x] Add entry to docs/changelog.rst
  - [x] Add stub and test

### DIFF
--- a/mitogen/buildah.py
+++ b/mitogen/buildah.py
@@ -1,0 +1,73 @@
+# Copyright 2019, David Wilson
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors
+# may be used to endorse or promote products derived from this software without
+# specific prior written permission.
+#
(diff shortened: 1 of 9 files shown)


mitogen/buildah.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2019, David Wilson
```
jordemort: This is literally just a slightly modified version of the Docker plugin so I just left your copyright.

jordemort:
Note that current versions of buildah have a bug in handling stdin on `buildah run` and you may need a version of buildah with https://github.com/containers/buildah/pull/1653 applied if you want to actually try this out.

jordemort:
Not sure what's up with CI here, failure looks unrelated.

dw:
Sorry for the delay, I have e-mails for the project getting archived just now :)

The CI failures are probably the Python 2.4 job which is missing a fork safety hook in one place, or the Ansible jobs which fail due to a race in the su method which is fixed on the async branch. Sorry about that

This looks perfect, thanks a ton!

dw:
Regarding the stdin issue, I'll try things out locally. There are quite a few knobs available to work around dodgy tooling, it's possible we already have a workaround for the problem in Buildah, but if not, we could always add one if it looks straightforward

jordemort:
@dw The PR to resolve my stdin handling issue (and a subsequent one that resolved even more) have been merged to buildah master (though they don't look like they have, since they don't actually use the GitHub merge function in their workflow) and should be in their 1.8.4 release.

Issue #482, opened 2019-01-25
Zombie process test


dw:
Following on from #480, this is a relatively frequently occurring problem, and it's always discovered by accident. Test suites should be updated to snapshot process list before / after test run, and fail if any stragglers exist.

- [x] Snapshot process list during setUpClass(), verify during teardownClass().
- [x] Snapshot process list before ./run_tests, ./mitogen_ansible_playbook etc, verify on complete
- [x] Snapshot target Docker container process list before / after ansible_tests run

Subprocess shutdown is asynchronous with regard to `broker.join()` right now, so at least initially this may need some nasty sleeps or retries to give the child a chance to go away on its own. It also bumps against #169 where the test suite does not reap processes reliably until process exit.

This may need to be done only for 'python' exes and 'mitogen:' comms, because e.g. DebOps may cause services to start in the container, and running tests on desktop while e.g. GNOME is doing stuff may cause flakeyness

Interesting comms/exes:

- [ ] python
- [ ] ssh
- [ ] mitogen:
- [ ] sudo
- [ ] su

cloudnull:
Maybe related - https://github.com/dw/mitogen/issues/510#issuecomment-459937284

The processes were not zombies, but I did have a situation were there were hundreds of mitogen processes still up on a machine after a lot of runs. That said, this all could have been due to something I was hacking on. I need to circle back and see if I can make it happen again.

dw:
I have gotten a bit sidetracked in the past week, but I found another case where they couldn't shut down due to the disconnect propagation work due in 0.2.4. The most recent change may be on master already, or still in my (currently very unhealthy) dmw branch.

This needs a more general fix. The problem in both recent cases was a crash during graceful shutdown. That procedure was historically bulletproof up to I think 0.2.3, and has gone downhill due to the disconnect propagation work.

IIRC what was happening last time was that the service thread pool was started up in the child, and that is what held the process alive after the broker thread crashed. What is supposed to happen is that as a final mea culpa, the broker thread kills the process right at the end of processing, to ensure absolutely nothing can leave anything behind. This might just need some blanket try/excepts in the appropriate places to ensure any broken callback or Stream.on_shutdown() cannot interrupt the shutdown process

PR #607, opened 2019-07-31
Update tested versions

sfermigier:
According to `tox.ini`, tested versions are: 2.6, 2.7 and 3.5, 3.6, 3.7.

Is 2.4 still supported BTW ?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -400,7 +400,7 @@ suitable for managing a fleet of potentially ancient corporate hardware, such
 as Red Hat Enterprise Linux 5, released in 2007.
 
 Every combination of Python 3.x/2.x parent and child should be possible,
-however at present only Python 2.4, 2.6, 2.7 and 3.6 are tested automatically.
+however at present only Python 2.6, 2.7 and 3.5, 3.6, 3.7 are tested automatically.
 
 
 Zero Dependencies


dw:
2.4 is actively tested, it's on Travis. tox.ini is mostly the 'property' of @moreati. I struggle to keep a working config of older versions of virtualenv matched with older Pythons, so I generally avoid tox usage and just run debug Python builds directly

moreati:
It was hard enough getting Python 2.6 working in Tox, the last time I tried. I'll take a look now, but if I can't do anything then I propose removing tox.ini.

dw:
I'm happy with removal, if it's not being used. Recently tried to use Tox to get virtualenvs for every sensible version of Ansible scripted, but ended up hand-writing something yet again out of frustration. Somehow I've never managed to grow an affinity for it

There are plenty of configs that Tox can't cover anyway (e.g. for me, OS X).. so in practice aside from ./run_tests under whatever venv I have to have active at that moment, I'm relying primarily on feedback from Travis/Azure during normal development already

moreati:
I'm exploring options for Tox + Docker. So far I've

- Made https://cloud.docker.com/repository/docker/moreati/python24, Python 2.4 in a container 
- Found https://hub.docker.com/r/shimizukawa/python-all, Pythons 2.4-3.7 in a container
- Found https://hub.docker.com/r/themattrix/tox/, Python 2.6-3.7 in a tox ready container (I think)

dw:
Running Pythons from a container could definitely work - I think the only reason it wasn't tried is because the tests start containers of their own, so some trickery is involved to re-export the same dockerd into the container.

There would be another interesting benefit in doing so -- I think it could potentially make testing for zombie processes after the test run completes (#482) much easier, since almost the same script could be used for checking for zombies inside a 'test target' container as in the 'test host' container

moreati:
### shimizukawa/python-all
Docker image based on Ubuntu (12.04, 14.04 or 18.04) + [deadsnakes](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa)<sup>1</sup>. Deadsnakes provides python interpreters, it doesn't provide PyPI packages such as virtualenv, pip etc. 

### themattrix/tox
Docker image based on (indirectly) Ubuntu 16.04 + [pyenv](https://github.com/pyenv/pyenv). pyenv installs Python interpreters by downloading, patching and building source code releases - along with setuptools and pip. It may even work on Mac.

### Versions
OS distro versions Shipped versions

Distro | setuptools | pip | virtualenv | tox
-- | -- | -- | -- | --
Ubuntu 12.04 | 0.6.24 | 1.0 | 1.7.1.2 | na
Ubuntu 14.04 | 3.3 | 1.5.4 | 1.11 | 1.6.0
Ubuntu 18.04<sup>2</sup> | 39.0.1 | 9.0.1 | 15.1.0 | 2.5.0

Last version with support

Python | setuptools | pip | virtualenv | tox
-- | -- | -- | -- | --
Python 2.4.6 | 1.4.2 | 1.1<sup>3</sup> | 1.7.2 | 1.3
Python 2.5.6 | 1.4.2 | 1.3.1<sup>3</sup> | 1.9.1 | 1.6.1
Python 2.6.9 | 36.8 | 9.0.3<sup>3</sup> | 15.2.0 | 2.9.1

Notes

1. Deadsnakes is a private package archive for CPython 2.3 - 3.current. I use it on my laptop.
2. Even though Ubuntu 18.04 ships tox 2.5.0, it cannot create Python 2.6 virtualenvs. The cause appears to be that Python 3.x is used to run tox, and this causes pip not to use a bundled copy of ordereddict - which is necessary with Python 2.6.
3. pip releases prior to 10.0.1 default to pypi.python.org as their index server, which no longer responds. The following in `~/.pypirc` coaxes them to use pypi.org.

       [distutils]
       index-servers = pypi

       [pypi]
       repository=https://pypi.org/legacy

dw:
A pyenv-based build is intriguing (assuming lots of caching was used!), as it could potentially mean a uniform way of getting debug CPython builds everywhere. The debug-enabled interpreters are a lot slower, but knowing it's always possible to get a backtrace out of a stuck test might be worth it. That could even open the door to automatically backtracing stuck CI jobs, which would be insanely useful

moreati:
A prototype for a tox replacement https://github.com/moreati/mitogen/tree/proto-yox. Python 2.4 only for now, based on pyenv

moreati:
proto-yox now caches downloads, and includes code coverage, it's still Python 2.4 only. I think it shows promise as a tox + deadsnakes replacement. It could be integrated with the existing `.ci/*` infrastructure. David, since you'd be the primary user, I'd like to check with you before I proceed further

1. Is this something you would use, and find useful? (or am I just creating more technical debt)
2. Is pyenv a viable way for you to install multiple pythons on macOS?
3. proto-yox could continue as a bash script, or become a Python script, or a Makefile. Do you have a preference?

moreati:
@dw ping. Any thoughts on those questions?

moreati:
Summary so far
- The official Mitogen tests are those that run on Travis and Azure Pipelines. If they don't pass then we don't merge.
- Tox tests are a convenience, used mainly by me. They cover fewer Python versions than the official tests.
- proto-yox is currently abandoned.
- My long term goal is to unify more of the official and tox testing code.

PR #608, opened 2019-07-31
Fix for sample in doc

sfermigier:
`log_to_file()` expects a filename.


dw:
I'd possibly recommend skipping the getting started guide entirely, and just focusing on the working code under examples/. "the_basics.py" is very heavily commented

dw:
Merged to docs-master. Thanks!

PR #612, opened 2019-08-08
Some smaller fixes

marc1006:


dw:
Thanks for these! Are you hitting them in an Ansible job, or some other program?  I'm a little concerned I didn't find them before merging that branch

The pkgutil change is curious. Where are you seeing this syntax error? It is shipped primarily for ModuleResponder on Python 2.4/2.5, I would not expect it to even be parsed on more recent versions.. Due to the intended target version, we can't use "as" there, it must be "sys.exc_info()[1]" if full 2->3 compatibility is desired

marc1006:
I'm using Mitogen for some tests - nothing related to Ansible.

Regarding the pkgutil change: as far as I can remember, I saw it a few months ago, but I'm not sure about that. I probably just saw it at the code and changed it. So feel free to drop it.

dw:
Wonderful :) I will merge as-is and figure it out from there.

dw:
Note current master completely changed the way logging works -- it is not set in stone yet, so if you don't like it or you bump into any other issues, please let me know. The upshot of the new approach is that logging package's filtering works the way it's supposed to, including for logs from remotes

marc1006:
Thanks a ton!

marc1006:
Will do :)

PR #683, opened 2020-01-19
Get mitogen.fakessh module working again

ryanofsky:
Fixes include

- Setting cloexec flag on pipe files, using set_inheritable on sockets,
  and close_fds=False on subprocess.Popen to work around file
  descriptors not being inheritable by default in new versions of python

- Adding mitogen.exit_status variable and avoiding os.kill call so fake
  'ssh' script is able to exit cleanly with correct status code

- Fixing broken os.dup call in ExternalContext._setup_master when input
  and output streams have the same descriptor

- Updating fakessh module to do necessary python3 string/byte
  conversions, and use updated mitogen Protocol, Stream, and Router apis

- Simplifying fakessh startup sequence so there aren't unnecessary
  differences between ways control and data handles are passed, and ways
  master and slave processes are initialized

- Fixing shutdown race conditions where subprocess exit handling or
  stdin EOF handling could result in a truncated stdout stream

- Updating and adding a lot of docstrings and comments

- Adding Process.proc is None / is not None assertions to be clear about
  which parts of fakessh.Process code are specific to the slave process,
  and which parts are specific to the master process.

- Re-enabling unit test case and updating an outdated file path so it
  passes

### DIFF
--- a/mitogen/core.py
+++ b/mitogen/core.py
@@ -1661,7 +1661,7 @@ class Stream(object):
         self.protocol = protocol
         self.protocol.stream = self
 
-    def accept(self, rfp, wfp):
+    def accept(self, rfp, wfp, cloexec=True):
         """
         Attach a pair of file objects to :attr:`receive_side` and
         :attr:`transmit_side`, after wrapping them in :class:`Side` instances.
@@ -1677,8 +1677,8 @@ class Stream(object):
         :param file wfp:
             The file object to transmit to.
         """
-        self.receive_side = Side(self, rfp)
-        self.transmit_side = Side(self, wfp)
+        self.receive_side = Side(self, rfp, cloexec=cloexec)
+        self.transmit_side = Side(self, wfp, cloexec=cloexec)
 
     def __repr__(self):
         return "<Stream %s #%04x>" % (self.name, id(self) & 0xffff,)
@@ -3767,7 +3767,7 @@ class ExternalContext(object):
         self.config = config
 
     def _on_broker_exit(self):
-        if not self.config['profiling']:
+        if not self.config['profiling'] and not hasattr(mitogen, "exit_status"):
             os.kill(os.getpid(), signal.SIGTERM)
 
     def _on_shutdown_msg(self, msg):
@@ -3822,9 +3822,11 @@ class ExternalContext(object):
 
         in_fd = self.config.get('in_fd', 100)
         in_fp = os.fdopen(os.dup(in_fd), 'rb', 0)
+        out_fp = os.fdopen(os.dup(self.config.get('out_fd', 1)), 'wb', 0)
+        # Avoid closing in_fd until after duplicating out_fd in case
+        # (in_fd == out_fd) are the same bidirectional socket fd
         os.close(in_fd)
 
-        out_fp = os.fdopen(os.dup(self.config.get('out_fd', 1)), 'wb', 0)
         self.stream = MitogenProtocol.build_stream(
             self.router,
             parent_id,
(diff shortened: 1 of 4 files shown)


mitogen/core.py
```diff
@@ -3767,7 +3767,7 @@ def __init__(self, config):
         self.config = config
 
     def _on_broker_exit(self):
-        if not self.config['profiling']:
+        if not self.config['profiling'] and not hasattr(mitogen, "exit_status"):
             os.kill(os.getpid(), signal.SIGTERM)
```
ryanofsky: This part of the change is hacky and was meant to be temporary. I'd appreciate any suggestions on replacing it.

I think I will want to the replace the `mitogen.exit_status` variable with a less general `fakessh.exit_status` variable, and add a separate `self.config['term_on_exit']`option to control whether SIGTERM is sent when the broker shuts down.

moreati:
@ryanofsky please could you update this to current master. Otherwise, may I update this branch on your behalf?

ryanofsky:
> @ryanofsky please could you update this to current master. Otherwise, may I update this branch on your behalf?

Hi! I think I can update the branch in the next few days or so, but if you're interested to do it sooner or take this over, it'd be more than welcome. Happy to add permissions to my github branch if that would help.

ryanofsky:
Hmm, apparently it was just a release notes documentation conflict so I used the github web interface to resolve it. The merge seems ok, but I can update if you want a different commit structure (maybe a rebase), or if something isn't working.

ryanofsky:
It seems that while this PR does restore fakessh functionality, taking it from a completely broken state to an updated and working state, there are still problems in CI.

One way to move forward with this PR might be to merge the code changes but keep the `@unittest2.skip('broken')` test annotation so fakessh can still be used while CI issues are debugged. fakessh is working well for me with the current PR and the `fakessh_test.py` test does pass reliably for me locally.

Looking at logs there seem to be 3 distinct CI issues:

1. [`ValueError: I/O operation on closed file`](https://dev.azure.com/mitogen-hq/mitogen/_build/results?buildId=66&view=logs&j=e324bded-a31e-5669-6941-770a6ec706f2&t=97708388-cdf6-5e51-bb90-3c79f75a8875&l=1870) exception causing [`self.assertEqual(return_code, 0)`](https://dev.azure.com/mitogen-hq/mitogen/_build/results?buildId=66&view=logs&j=e324bded-a31e-5669-6941-770a6ec706f2&t=97708388-cdf6-5e51-bb90-3c79f75a8875&l=2627) test failure. This seems to be a race condition that doesn't always happen. When it does happen rsync actually succeeds but there is an error closing down connections.

2. [`AttributeError: '_socket.socket' object has no attribute 'set_inheritable'`](https://dev.azure.com/mitogen-hq/mitogen/_build/results?buildId=66&view=logs&j=364fd14b-95c5-5834-280a-1629ce194374&t=1b0abb5f-92e2-5ce1-22fe-ad8871c3092d&l=789) error in python2, because the method only exists in python3. It should be pretty easy to fix because the call isn't necessary in python2 (descriptors are inherited by default there).

3. [`assert [proc.wait() for proc in procs] == [0] * len(procs)`](https://dev.azure.com/mitogen-hq/mitogen/_build/results?buildId=66&view=logs&j=39882f3a-9103-5f46-8b73-c9b5dbece82f&t=b15853ad-39ae-51f8-ef8e-7d055aaf8ba7&l=133) in ansible install. It's not actually clear that this is related to the PR, but it might be.

mhartmay:
Any updates? What’s needed to get this merged?

ryanofsky:
> Any updates? What’s needed to get this merged?

There were some minor conflicts with the master branch, so I just rebased this including @moreati's patch and conflict resolution from #924.

I'm not currently using or relying on this functionality and I don't think I have time to make major updates, so would be happy if someone wants to take this over. But also happy to rebase and apply any fixes or suggestions here.

Issue #568, opened 2019-03-19
Make strategy_plugins obsolete


varac:
I'm super impressed by the speedup of ansible_mitogen !
However, there's one thing blocking the rollout in our opensource project, and this is the fact that you need to specify the location of the ansible_mitogen strategy plugin in `ansible.cfg`.
The location will veary depending of the type of installation, i.e. some will install it under `~/.local/lib/python3.6`, some under `/user/local/lib/python3.7`, some in a virtualenv etc.

We like to ship a `ansibile.cfg` that doesn't need to get modified manually, so I wonder if there's any change that ansible will try to find the plugin path by itself ? That would make it even easier for the user.

dw:
Hi :)  The simplest solution is to avoid installing via pip. The installation steps avoid pip for this very reason -- the package directory can be placed in any path (e.g. in the same dir as ansible.cfg), and fixes the Python path up internally to cope with the install location.

Can't you ship the extension and say, place it in the same directory as ansible.cfg? That way, 'strategy_plugins = mitogen-1.2.3/ansible_mitogen/plugins/strategy' "should" always succeed.

There are a bunch more hooks where it might be possible to trick it on to the Ansible loader path, but I'd need to understand more about your project first in order to suggest one

varac:
Sure, we can do that, it's just that we already ship a `requirements.txt` with multiple python deps that we install via pip, and therefore it feel weird to ship a bundled version of one exceptional python module. We'd like to separate our code from external modules.

dw:
Totes understand the desire to use pip :) The trouble is that Ansible's plugin loader paths are completely separate.

There is at least one interpreter-level hack that is possible to cure this "once and for all", but making global changes to the interpreter config is a very nasty thing for any Python package to do.

varac:
Ok, it's not a big deal, I wanted to check if this would be easily solveable. We'll probably ship ansible_mitogen for now, until this has been implemented. Thx for the quick response!

dw:
Thanks for reporting it anyway -- this is a known usability issue and there wasn't an existing ticket for it

dmsimard:
@dw @varac we had similar challenges getting users to configure the ARA callback plugin :)

Once upon a time, we had users do something like this:
```
export ara_location=$(python -c "import os,ara; print(os.path.dirname(ara.__file__))")
export ANSIBLE_CALLBACK_PLUGINS=$ara_location/plugins/callbacks
```
In a script it's not too bad but we needed something better for humans.

[...]

PR #685, opened 2020-01-26
Adds more dynamic installation

ssbarnea:
This enables a cli that can return the path of the plugins, enabling
a more dynamic installation method of mitogen, one that does not
require changing ansible.cfg files.


dw:
There is some related discussion on #568. I'm -1 on installing any command line scripts, but +1 on having something reachable via "python -m". It's not as pretty, but it means not having a little sidecar polluting the path and requiring setuptools to install.

Also since this is Ansible extension-specific (likely the extension will move to its own repo soon), can it please live in ansible_mitogen

PR #702, opened 2020-03-20
[BUGFIX] Respect SSH host key checking settings from host

thommyhh:
Use the host or group specific settings regarding SSH host key checking,
so the checking can be disabled for certain hosts or group of hosts.

### DIFF
--- a/ansible_mitogen/connection.py
+++ b/ansible_mitogen/connection.py
@@ -120,10 +120,15 @@ def _connect_ssh(spec):
     """
     Return ContextService arguments for an SSH connection.
     """
-    if C.HOST_KEY_CHECKING:
+    if spec.ssh_host_key_checking() == 'no' or spec.ssh_host_key_checking() is False:
+        check_host_keys = 'ignore'
+    elif spec.ssh_host_key_checking() == 'yes' or spec.ssh_host_key_checking() is True:
         check_host_keys = 'enforce'
     else:
-        check_host_keys = 'ignore'
+        if C.HOST_KEY_CHECKING:
+            check_host_keys = 'enforce'
+        else:
+            check_host_keys = 'ignore'
 
     # #334: tilde-expand private_key_file to avoid implementation difference
     # between Python and OpenSSH.
--- a/ansible_mitogen/transport_config.py
+++ b/ansible_mitogen/transport_config.py
@@ -197,6 +197,12 @@ class Spec(with_metaclass(abc.ABCMeta, object)):
         """
 
     @abc.abstractmethod
+    def ssh_host_key_checking(self):
+        """
+        Whether the SSH client should check host keys or not
+        """
+
+    @abc.abstractmethod
     def ssh_args(self):
         """
         The list of additional arguments that should be included in an SSH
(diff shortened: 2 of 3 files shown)


ansible_mitogen/transport_config.py
```diff
@@ -384,6 +390,12 @@ def ansible_ssh_timeout(self):
             self.timeout()
         )
 
+    def ssh_host_key_checking(self):
+        if self._connection.get_task_var('ansible_ssh_host_key_checking') is not None:
```
s1113950: Please simplify this to:
```
return self._connection.get_task_var('ansible_ssh_host_key_checking') or self._connection.get_task_var('ansible_host_key_checking')
```
which saves multiple calls to `get_task_var('ansible_ssh_host_key_checking')`

thommyhh: @s1113950 I tried your proposed change. The problem is, that if
* `ansible_ssh_host_key_checking` is set to false and
* `ansible_host_key_checking` is not set

the `or` version would return `None` instead of `false`.

s1113950: ahh! Good catch. Then instead of calling `self._connection.get_task_var('ansible_ssh_host_key_checking')` twice, can you make it a variable and use it in your if block? Reason being `get_task_var` can get expensive depending on how much walking-the-stack needs to be done

thommyhh: Good point. Tanks.

moreati:
@thommyhh please could you update this to current master. Otherwise, may I update this branch on your behalf?

thommyhh:
@moreati Please feel free to rebase to current master. I'm missing local setup currently.

ansible_mitogen/connection.py
```diff
@@ -120,10 +120,15 @@ def _connect_ssh(spec):
     """
     Return ContextService arguments for an SSH connection.
     """
-    if C.HOST_KEY_CHECKING:
+    if spec.ssh_host_key_checking() == 'no' or spec.ssh_host_key_checking() is False:
```
s1113950, line 123: Instead of potentially calling `spec.ssh_host_key_checking()` 4 times, please make this a var and then refer to it in your if statements:
```
ssh_host_key_checking = spec.ssh_host_key_checking()
if ssh_host_key_checking in ('no', False):
  check_host_keys = 'ignore'
elif ssh_host_key_checking in ('yes', True):
  check_host_keys = 'enforce'
else:
  ...
```

s1113950 (changes requested):

docs/changelog.rst
```diff
@@ -21,7 +21,12 @@ v0.2.10 (unreleased)
 To avail of fixes in an unreleased version, please download a ZIP file
 `directly from GitHub <https://github.com/dw/mitogen/>`_.
 
-*(no changes)*
+Mitogen for Ansible
+~~~~~~~~~~~~~~~~~~~
+* It is not possible to control SSH host key checking for certain host or group
```
moreati, line 26: Spelling s/not/now/

PR #710, opened 2020-04-17
Handles a `wait_for_connection` call right after a task caused a shutdown

### LINKED ISSUE #655: wait_for_connection gives errors
I'm using ansible + mitogen for automated server updates on Ubuntu. I'm using these tasks to check if a reboot is necessary after the apt upgrade task:

```
  - name: Check if reboot is required
    stat:
      path: /var/run/reboot-required
    register: reboot_required

  - name: Reboot server
    shell: sleep 2 && shutdown -r now "Ansible updates triggered"
    async: 1
    poll: 0
    when: reboot_required.stat.exists == True

  - name: Wait 300 seconds for server to become available
    wait_for_connection:
      delay: 30
      timeout: 300
    when: reboot_required.stat.exists == True
```

This can be manually reproduced by just creating a /var/run/reboot-required file.

I recently updated to the latest ansible (2.8.5) and mitogen (master branch). Previously the wait_for_connection would trigger a warning ([WARNING]: Reset is not implemented for this connection), but now the following error is shown every second, from the moment the reboot is complete:

```
ERROR! [mux  23628] 12:00:35.305036 E mitogen.service: Pool(1c50, size=32, th='mitogen.Pool.1c50.24'): while invoking u'reset' of u'ansible_mitogen.services.ContextService'
  File "/home/epartment/epaflex/mitogen-master/mitogen/service.py", line 621, in _on_service_call
    return invoker.invoke(method_name, kwargs, msg)
  File "/home/epartment/epaflex/mitogen-master/mitogen/service.py", line 307, in invoke
    response = self._invoke(method_name, kwargs, msg)
  File "/home/epartment/epaflex/mitogen-master/mitogen/service.py", line 293, in _invoke
    ret = method(**kwargs)
  File "/home/epartment/epaflex/mitogen-master/ansible_mitogen/services.py", line 186, in reset
    mitogen.core.listen(context, 'disconnect', l.put)
  File "/home/epartment/epaflex/mitogen-master/mitogen/core.py", line 431, in listen
    _signals(obj, name).append(func)
  File "/home/epartment/epaflex/mitogen-master/mitogen/core.py", line 421, in _signals
    obj.__dict__
AttributeError: 'NoneType' object has no attribute '__dict__'
```

When disabling mitogen, the playbook runs fine.

s1113950:
Fixes #655 

`wait_for_connection`  in Ansible calls both a `connection.reset()` as well as an `_execute_module`.

In the case of issue 655, a shutdown call right before the `wait_for_connection` caused Mitogen to disconnect from the host and then be stuck in a bad state when `_execute_module` was called.

`wait_for_connection` is set up to retry continually to connect, which was masking errors that Mitogen was throwing when it couldn't set up a connection. 

To fix I forced Mitogen to reconnect if the `ping` module is used with `wait_for_connection` and made sure that if we're trying to create a `Pool` that we have a `router`.

### DIFF
--- a/tests/ansible/regression/issue_655__wait_for_connection_error.yml
+++ b/tests/ansible/regression/issue_655__wait_for_connection_error.yml
@@ -1,0 +1,85 @@
+# https://github.com/dw/mitogen/issues/655
+# Spins up a Centos8 container and runs the wait_for_connection test inside of it
+# Doing it this way because the shutdown command causes issues in our tests
+#   since things are ran on localhost; Azure DevOps loses connection and fails
+# TODO: do we want to install docker a different way to be able to do this for other tests too
+---
+# this should only run on our Mac hosts
+- hosts: target
+  any_errors_fatal: True
+  gather_facts: yes
+  become: no
+  tasks:
+    - name: set up test container and run tests inside it
+      block:
+        - name: install deps
+          block:
+            - name: install docker
+              shell: |
+                # NOTE: for tracking purposes: https://github.com/docker/for-mac/issues/2359
+                # using docker for mac CI workaround: https://github.com/drud/ddev/pull/1748/files#diff-19288f650af2dabdf1dcc5b354d1f245
+                DOCKER_URL=https://download.docker.com/mac/stable/31259/Docker.dmg && 
+                curl -O -sSL $DOCKER_URL && 
+                open -W Docker.dmg && cp -r /Volumes/Docker/Docker.app /Applications
+                sudo /Applications/Docker.app/Contents/MacOS/Docker --quit-after-install --unattended && 
+                ln -s /Applications/Docker.app/Contents/Resources/bin/docker /usr/local/bin/docker && 
+                nohup /Applications/Docker.app/Contents/MacOS/Docker --unattended & 
+                # wait 2 min for docker to come up
+                counter=0 && 
+                  while ! /usr/local/bin/docker ps 2>/dev/null ; do
+                    if [ $counter -lt 24 ]; then
+                      let counter=counter+1
+                    else
+                      exit 1
+                    fi
+                    sleep 5
+                  done
+
+        # python bindings (docker_container) aren't working on this host, so gonna shell out
+        - name: create docker container
+          shell: /usr/local/bin/docker run --name testMitogen -d --rm centos:8 bash -c "sleep infinity & wait"
+
+        - name: add container to inventory
+          add_host:
+            name: testMitogen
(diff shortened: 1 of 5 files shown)


s1113950:
/aside to travis: oh travis! Please report results!

s1113950:
Travis does not seem to wanna launch a test 🤦

s1113950:
New plan:
don't use `docker-machine` since it seems to not work well with `add_host` (no docker sock file). Wonder if it's possible to install docker for mac via terminal only. Could also figure out how to connect docker-machine with an `add_host`-ed docker container setup.

s1113950:
@cognifloyd can you review this one as well? :) 🙏

s1113950:
ooh I'll squash commits into 1 actually since there's no current chat history

tests/ansible/regression/issue_655__wait_for_connection_error.yml
```diff
@@ -0,0 +1,85 @@
+# https://github.com/dw/mitogen/issues/655
+# Spins up a Centos8 container and runs the wait_for_connection test inside of it
+# Doing it this way because the shutdown command causes issues in our tests
+#   since things are ran on localhost; Azure DevOps loses connection and fails
+# TODO: do we want to install docker a different way to be able to do this for other tests too
+---
+# this should only run on our Mac hosts
```
cognifloyd, line 7: What prevents this from running on non-mac hosts? I don't see any safety checks to bail if the host is not a mac, instead it will just go up in smoke when the dmg can't be unpacked, or the /Volumes directory doesn't exist.

s1113950: The `target` host only runs on our Mac setup because Mac is the only test that runs `localhost_ansible` tests (see `azure-pipelines.yml`). Technically someone could run it themselves on a non-mac and it might not work but the tests are really set up for running through CI mostly so I didn't worry about that. I left a note so that if we end up running `localhost_ansible` on another host then we'd know to refactor this file a bit! It's not super clear though; if you'd like I could edit the comment to reflect something different!

luto:
fixes the `reboot` module as well. Thank you!

tests/ansible/regression/issue_655__wait_for_connection_error.yml
cognifloyd: Nah. Comment is enough. :)

s1113950: Sounds good!

PR #714, opened 2020-04-27
Correct latest ansible version supported

scottsb:
Self-explanatory. (Might it be worth removing this from the docs lest it continue to get stale?)

### DIFF
--- a/docs/ansible_detailed.rst
+++ b/docs/ansible_detailed.rst
@@ -145,7 +145,7 @@ Testimonials
 Noteworthy Differences
 ----------------------
 
-* Ansible 2.3-2.8 are supported along with Python 2.6, 2.7, 3.6 and 3.7. Verify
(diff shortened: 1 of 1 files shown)


s1113950:
Technically Ansible 2.9 isn't fully supported yet 😅 see https://github.com/dw/mitogen/pull/705 for more details. It works with some things but we weren't even running tests on it to verify that it worked. Additionally, things like collections aren't supported yet. The plan is after basic support lands to work on collections.

s1113950:
Base 2.9 support has been added, and collections are being worked on. I think after collections land this can land as well.

s1113950:
Thanks for the PR! Collections support will be done in Ansible 2.10 support, and enough people have been using Ansible 2.9 with Mitogen for some time now so I'll merge this in. I was told that collections support wasn't fully integrated until Ansible 2.10's recent release anyhow.

PR #715, opened 2020-05-08
Ansible 2.10 + Collections support

### LINKED ISSUE #652: Support for ansible collections import hook
Ansible 2.8.4 + mitogen 0.2.8
Collections installed in ~/.ansible/collections
Playbook and all roles are in a collection.

It looks like the mitogen linear_strategy is not compatible with the collections loader.

Sorry I'm working remotely (to test mitogen with this playbook) on my phone so all I can provide are screenshots of the error. I can look up more details next week.

Screenshot_20191005-191340~4

Screenshot_20191005-191941~2

### LINKED ISSUE #731: ansible 2.10 support
There doesn't seem to be an existing issue referencing ansible 2.10 issues yet, so here goes :)

```
ansible 2.10.0b1.post0
  config file = /export/home/orion-admin/ansible-boulder/ansible.cfg
  configured module search path = [u'/export/home/orion-admin/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /export/home/orion-admin/ansible/lib/ansible
  executable location = /export/home/orion-admin/ansible/bin/ansible
  python version = 2.7.5 (default, Apr  1 2020, 10:09:19) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]
Using /export/home/orion-admin/ansible-boulder/ansible.cfg as config file
Parsed /export/home/orion-admin/ansible-boulder/inventory/boulder.cobbler.yml inventory source with cobbler plugin
Skipping due to inventory source not ending in "cobbler.yaml" nor "cobbler.yml"
cobbler declined parsing /export/home/orion-admin/ansible-boulder/inventory/hosts as it did not pass its verify_file() method
Parsed /export/home/orion-admin/ansible-boulder/inventory/hosts inventory source with ini plugin
Parsed /export/home/orion-admin/ansible-boulder/inventory/monterey.cobbler.yml inventory source with cobbler plugin
Parsed /export/home/orion-admin/ansible-boulder/inventory/seattle.cobbler.yml inventory source with cobbler plugin
[top  4317] 10:19:00.884328 D ansible_mitogen.affinity: CPU mask for Ansible top-level process: 0x000002
[top  4317] 10:19:00.884631 D ansible_mitogen.process: inherited open file limits: soft=1024 hard=4096
[top  4317] 10:19:00.884730 D ansible_mitogen.process: raised soft open file limit from 1024 to 4096
[mux  4327] 10:19:00.887142 D ansible_mitogen.affinity: CPU mask for MuxProcess 0: 0x000001
[mux  4327] 10:19:00.902435 D mitogen.service: Pool(a110, size=32, th='MainThread'): initialized
[mux  4327] 10:19:00.902749 D ansible_mitogen.process: Service pool configured: size=32
META: ran handlers
[task 4361] 10:19:00.934268 D ansible_mitogen.affinity: CPU mask for WorkerProcess: 0x000004
The full traceback is:
  File "/export/home/orion-admin/ansible/lib/ansible/executor/task_executor.py", line 155, in run
    res = self._execute()
  File "/export/home/orion-admin/ansible/lib/ansible/executor/task_executor.py", line 610, in _execute
    self._handler = self._get_action_handler(connection=self._connection, templar=templar)
  File "/export/home/orion-admin/ansible/lib/ansible/executor/task_executor.py", line 1063, in _get_action_handler
    collection_list=collections
  File "/export/home/orion-admin/ansible-boulder/mitogen/ansible_mitogen/strategy.py", line 144, in wrap_action_loader__get
    return adorned_klass(*args, **kwargs)
  File "/export/home/orion-admin/ansible-boulder/mitogen/ansible_mitogen/mixins.py", line 118, in __init__
    super(ActionModuleMixin, self).__init__(task, connection, *args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'collection_list'
barry.cora.nwra.com | FAILED! => {
    "msg": "Unexpected failure during module execution.",
    "stdout": ""
}
[mux  4327] 10:19:00.980813 D mitogen: Waker(fd=6/8): disconnecting
```

s1113950:
Fixes #652 
Fixes #731 

Summary of changes:
1. Rename some monkeypatches because Ansible 2.10 changed how stuff is called (for this reason I think we should stick with 0.3.0 being backwards-incompatible and do a separate 0.2.10 release for everything currently in `master`)
2. Hack the `ActionModuleMixin` base class because collection support didn't have an ActionBase for some reason. This hopefully can be removed because I initially added it in when I was still on Ansible's devel branch, pre-2.10.
3. Use Ansible's collections logic to determine where collections are installed, and then tell the `master` where they are via its `sys.path` so it can load them later when it looks for where dependencies live
4. Handle the fake module `__synthetic__` thing that Ansible added to collections modules

TODO:
- ~[ ] Synchronize module is broken, it moved from an ansible action plugin here: https://github.com/ansible/ansible/blob/v2.9.13/lib/ansible/plugins/action/synchronize.py to being in the builtin ansible collection `posix`: `redirecting (type: action) ansible.builtin.synchronize to ansible.posix.synchronize`~ -> https://github.com/dw/mitogen/pull/715#issuecomment-702518017
- [x] Switching between `mitogen_linear` and `linear` strategies isn't working, something about Ansible 2.10 changed how strategy is loaded. Doing so in the same play file in different `hosts` blocks is failing (https://github.com/dw/mitogen/blob/master/tests/ansible/integration/strategy/mixed_vanilla_mitogen.yml#L2 this test, which can be reproduced locally as well). ~Not sure if this is a bug with Ansible 2.10 or with Mitogen specifically yet.~ It's an issue with how Mitogen overrides the `connection` obj (this works in Ansible 2.10 when I tried switching from `linear` to `free` and back again), here's the traceback:
```
"The full traceback is:",
        "Traceback (most recent call last):",
        "  File \"/mitogen-test/.venv3/lib/python3.8/site-packages/ansible/executor/task_executor.py\", line 158, in run",
        "    res = self._execute()",
        "  File \"/mitogen-test/.venv3/lib/python3.8/site-packages/ansible/executor/task_executor.py\", line 663, in _execute",
        "    result = self._handler.run(task_vars=variables)",
        "  File \"/mitogen-test/.venv3/lib/python3.8/site-packages/ansible/plugins/action/normal.py\", line 46, in run",
        "    result = merge_hash(result, self._execute_module(task_vars=task_vars, wrap_async=wrap_async))",
        "  File \"/mitogen-test/.venv3/lib/python3.8/site-packages/ansible/plugins/action/__init__.py\", line 826, in _execute_module",
        "    self._make_tmp_path()",
        "  File \"/mitogen-test/.venv3/lib/python3.8/site-packages/ansible/plugins/action/__init__.py\", line 388, in _make_tmp_path",
        "    tmpdir = self._remote_expand_user(self.get_shell_option('remote_tmp', default='~/.ansible/tmp'), sudoable=False)",
        "  File \"/mitogen-test/.venv3/lib/python3.8/site-packages/ansible/plugins/action/__init__.py\", line 709, in _remote_expand_user",
        "    data = self._low_level_execute_command(cmd, sudoable=False)",
        "  File \"/mitogen-test/.venv3/lib/python3.8/site-packages/ansible/plugins/action/__init__.py\", line 1121, in _low_level_execute_command",
        "    rc, stdout, stderr = self._connection.exec_command(cmd, in_data=in_data, sudoable=sudoable)",
        "  File \"/private/tmp/mitogen/ansible_mitogen/connection.py\", line 996, in exec_command",
        "    rc, stdout, stderr = self.get_chain().call(",
        "  File \"/private/tmp/mitogen/ansible_mitogen/connection.py\", line 942, in get_chain",
        "    self._connect()",
        "  File \"/private/tmp/mitogen/ansible_mitogen/connection.py\", line 838, in _connect",
        "    inventory_name, stack = self._build_stack()",
        "  File \"/private/tmp/mitogen/ansible_mitogen/connection.py\", line 774, in _build_stack",
        "    inventory_name=self.get_task_var('inventory_hostname'),",
        "  File \"/private/tmp/mitogen/ansible_mitogen/connection.py\", line 649, in get_task_var",
        "    task_vars = self._get_task_vars()",
        "  File \"/private/tmp/mitogen/ansible_mitogen/connection.py\", line 598, in _get_task_vars",
        "    for line in traceback.format_stack():",
        "NameError: name 'traceback' is not defined",
```
possibly broken from here?: https://github.com/ansible/ansible/commit/8c2754e6d366a8e7c47976717193262b432fe83e , `self._connection` is set to the old Mitogen connection obj, despite `strategy` being set to `linear`
- [x] Peg to ansible 2.10 now that it's released, rather than cloning from github
- [x] Fix up the tests to only test 2.10 versions of stuff
- [x] Add support for collections installed to site-packages (what happens with `pip install ansible==2.10.0`) instead of via `ansible-galaxy collection install`
- [x] fix a Debian stretch (oldstable) connection error
- [ ] Add documentation explaining the changes
- [ ] Remove `NOTE` sections in the code where backwards compat breaks, since it's been decided it's gonna break
- ~[ ] Add more test cases around different common collection usages~ test cases use installed collections in them, so I think we're covered here
- ~[ ] Anything else that Ansible 2.10 might have broken~ fixed everything that's been reported so far, and will fix future things in future releases
- ~[ ] get debops working: https://github.com/dw/mitogen/pull/715#issuecomment-690868413 it might not be 2.10 compatible just yet for our use case~ -> disabling debops because ran into https://github.com/debops/debops/issues/1521
- [x] See if collections processing can be done more efficiently

### DIFF
--- a/.ci/ansible_install.py
+++ b/.ci/ansible_install.py
@@ -11,10 +11,8 @@ batches = [
         'pip install '
             '-r tests/requirements.txt '
             '-r tests/ansible/requirements.txt',
-        # 'pip install -q ansible=={0}'.format(ci_lib.ANSIBLE_VERSION)
-        # ansible v2.10 isn't out yet so we're installing from github for now
         # encoding is required for installing ansible 2.10 with pip2, otherwise we get a UnicodeDecode error
-        'LC_CTYPE=en_US.UTF-8 LANG=en_US.UTF-8 pip install -q {}'.format(ci_lib.ANSIBLE_VERSION)
+        'LC_CTYPE=en_US.UTF-8 LANG=en_US.UTF-8 pip install -q ansible=={0}'.format(ci_lib.ANSIBLE_VERSION)
     ]
 ]
 
@@ -24,8 +22,3 @@ batches.extend(
 )
 
 ci_lib.run_batches(batches)
-
-# after ansible is installed, install common collections until ansible==2.10 comes out
-ci_lib.run('ansible-galaxy collection install community.general')
-ci_lib.run('ansible-galaxy collection install ansible.netcommon')
-ci_lib.run('ansible-galaxy collection install ansible.posix')
--- a/.ci/localhost_ansible_tests.py
+++ b/.ci/localhost_ansible_tests.py
@@ -46,5 +46,5 @@ with ci_lib.Fold('machine_prep'):
 with ci_lib.Fold('ansible'):
     os.chdir(TESTS_DIR)
     playbook = os.environ.get('PLAYBOOK', 'all.yml')
-    run('./run_ansible_playbook.py %s -l target %s',
+    run('./run_ansible_playbook.py %s -l target %s -vvv',
         playbook, ' '.join(sys.argv[1:]))
--- a/.ci/localhost_ansible_tests.py
+++ b/.ci/localhost_ansible_tests.py
@@ -21,8 +21,14 @@ with ci_lib.Fold('unit_tests'):
 
 with ci_lib.Fold('job_setup'):
     os.chmod(KEY_PATH, int('0600', 8))
+    # NOTE: sshpass v1.06 causes errors so pegging to 1.05 -> "msg": "Error when changing password","out": "passwd: DS error: eDSAuthFailed\n", 
+    # there's a checksum error with "brew install http://git.io/sshpass.rb" though, so installing manually
     if not ci_lib.exists_in_path('sshpass'):
-        run("brew install esolitos/ipa/sshpass")
+        run("curl -O -L  https://sourceforge.net/projects/sshpass/files/sshpass/1.05/sshpass-1.05.tar.gz && \
+                tar xvf sshpass-1.05.tar.gz && \
+                cd sshpass-1.05 && \
+                ./configure && \
+                sudo make install")
 
 
 with ci_lib.Fold('machine_prep'):
--- a/.ci/localhost_ansible_tests.py
+++ b/.ci/localhost_ansible_tests.py
@@ -22,7 +22,7 @@ with ci_lib.Fold('unit_tests'):
 with ci_lib.Fold('job_setup'):
     os.chmod(KEY_PATH, int('0600', 8))
     if not ci_lib.exists_in_path('sshpass'):
-        run("brew install http://git.io/sshpass.rb")
+        run("brew install esolitos/ipa/sshpass")
 
 
 with ci_lib.Fold('machine_prep'):
--- a/.ci/localhost_ansible_tests.py
+++ b/.ci/localhost_ansible_tests.py
@@ -24,7 +24,7 @@ with ci_lib.Fold('job_setup'):
     # NOTE: sshpass v1.06 causes errors so pegging to 1.05 -> "msg": "Error when changing password","out": "passwd: DS error: eDSAuthFailed\n", 
     # there's a checksum error with "brew install http://git.io/sshpass.rb" though, so installing manually
     if not ci_lib.exists_in_path('sshpass'):
-        run("curl -O -L  https://sourceforge.net/projects/sshpass/files/sshpass/1.05/sshpass-1.05.tar.gz && \
+        os.system("curl -O -L  https://sourceforge.net/projects/sshpass/files/sshpass/1.05/sshpass-1.05.tar.gz && \
                 tar xvf sshpass-1.05.tar.gz && \
                 cd sshpass-1.05 && \
                 ./configure && \
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,17 +53,17 @@ matrix:
     - python: "2.7"
       env: MODE=ansible VER=2.10.0
     # 2.10 -> {debian, centos6, centos7}
-    - python: "2.6"
-      env: MODE=ansible VER=2.10.0
+    # - python: "2.6"
+    #   env: MODE=ansible VER=2.10.0
 
     # 2.10 -> {centos5}
-    - python: "2.6"
-      env: MODE=ansible DISTROS=centos5 VER=2.10.0
+    # - python: "2.6"
+    #   env: MODE=ansible DISTROS=centos5 VER=2.10.0
 
     # Mitogen tests.
     # 2.4 -> 2.4
-    - language: c
-      env: MODE=mitogen_py24 DISTROS=centos5 VER=2.10.0
+    # - language: c
+    #   env: MODE=mitogen_py24 DISTROS=centos5 VER=2.10.0
     # 2.7 -> 2.7 -- moved to Azure
     # 2.7 -> 2.6
     #- python: "2.7"
@@ -71,9 +71,9 @@ matrix:
     - python: "3.6"
       env: MODE=mitogen DISTROS=centos7 VER=2.10.0
     # 2.6 -> 2.7
-    - python: "2.6"
-      env: MODE=mitogen DISTROS=centos7 VER=2.10.0
+    # - python: "2.6"
+    #   env: MODE=mitogen DISTROS=centos7 VER=2.10.0
     # 2.6 -> 3.5
-    - python: "2.6"
-      env: MODE=mitogen DISTROS=debian-py3 VER=2.10.0
+    # - python: "2.6"
+    #   env: MODE=mitogen DISTROS=debian-py3 VER=2.10.0
     # 3.6 -> 2.6 -- moved to Azure
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,8 +42,9 @@ matrix:
     #   env: MODE=debops_common VER=2.10.0
 
     # Sanity check against vanilla Ansible. One job suffices.
-    - python: "2.7"
-      env: MODE=ansible VER=2.10.0 DISTROS=debian STRATEGY=linear
+    # https://github.com/dw/mitogen/pull/715#issuecomment-719266420 migrating to Azure for now
+    # - python: "2.7"
+    #   env: MODE=ansible VER=2.10.0 DISTROS=debian STRATEGY=linear
 
     # ansible_mitogen tests.
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,12 +21,11 @@ install:
 - pip install -U pip==20.2.1
 - .ci/${MODE}_install.py
 
+# Travis has a 4MB log limit (https://github.com/travis-ci/travis-ci/issues/1382), but verbose Mitogen logs run larger than that
+# in order to keep verbosity to debug a build failure, will run with this workaround: https://stackoverflow.com/a/26082445
 script:
 - .ci/spawn_reverse_shell.py
 - MODE=${MODE} .ci/travis.sh
-
-# Travis has a 4MB log limit (https://github.com/travis-ci/travis-ci/issues/1382), but verbose Mitogen logs run larger than that
-# in order to keep verbosity to debug a build failure, will run with this workaround: https://stackoverflow.com/a/26082445
 
 # To avoid matrix explosion, just test against oldest->newest and
 # newest->oldest in various configuartions.
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,45 +34,45 @@ matrix:
     # Debops tests.
     # 2.10; 3.6 -> 2.7
     - python: "3.6"
-      env: MODE=debops_common VER=git+https://github.com/ansible/ansible.git@v2.10.0
+      env: MODE=debops_common VER=2.10.0
     # 2.10; 2.7 -> 2.7
     - python: "2.7"
-      env: MODE=debops_common VER=git+https://github.com/ansible/ansible.git@v2.10.0
+      env: MODE=debops_common VER=2.10.0
 
     # Sanity check against vanilla Ansible. One job suffices.
     - python: "2.7"
-      env: MODE=ansible VER=git+https://github.com/ansible/ansible.git@v2.10.0 DISTROS=debian STRATEGY=linear
+      env: MODE=ansible VER=2.10.0 DISTROS=debian STRATEGY=linear
 
     # ansible_mitogen tests.
 
     # 2.10 -> {debian, centos6, centos7}
     - python: "3.6"
-      env: MODE=ansible VER=git+https://github.com/ansible/ansible.git@v2.10.0
+      env: MODE=ansible VER=2.10.0
     # 2.10 -> {debian, centos6, centos7}
     - python: "2.7"
-      env: MODE=ansible VER=git+https://github.com/ansible/ansible.git@v2.10.0
+      env: MODE=ansible VER=2.10.0
     # 2.10 -> {debian, centos6, centos7}
     - python: "2.6"
-      env: MODE=ansible VER=git+https://github.com/ansible/ansible.git@v2.10.0
+      env: MODE=ansible VER=2.10.0
 
     # 2.10 -> {centos5}
     - python: "2.6"
-      env: MODE=ansible DISTROS=centos5 VER=git+https://github.com/ansible/ansible.git@v2.10.0
+      env: MODE=ansible DISTROS=centos5 VER=2.10.0
 
     # Mitogen tests.
     # 2.4 -> 2.4
     - language: c
-      env: MODE=mitogen_py24 DISTROS=centos5 VER=git+https://github.com/ansible/ansible.git@v2.10.0
+      env: MODE=mitogen_py24 DISTROS=centos5 VER=2.10.0
     # 2.7 -> 2.7 -- moved to Azure
     # 2.7 -> 2.6
     #- python: "2.7"
       #env: MODE=mitogen DISTRO=centos6
     - python: "3.6"
-      env: MODE=mitogen DISTROS=centos7 VER=git+https://github.com/ansible/ansible.git@v2.10.0
+      env: MODE=mitogen DISTROS=centos7 VER=2.10.0
     # 2.6 -> 2.7
     - python: "2.6"
-      env: MODE=mitogen DISTROS=centos7 VER=git+https://github.com/ansible/ansible.git@v2.10.0
+      env: MODE=mitogen DISTROS=centos7 VER=2.10.0
     # 2.6 -> 3.5
     - python: "2.6"
-      env: MODE=mitogen DISTROS=debian-py3 VER=git+https://github.com/ansible/ansible.git@v2.10.0
+      env: MODE=mitogen DISTROS=debian-py3 VER=2.10.0
     # 3.6 -> 2.6 -- moved to Azure
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,8 @@ matrix:
     #   env: MODE=debops_common VER=2.10.0
 
     # Sanity check against vanilla Ansible. One job suffices.
-    # https://github.com/dw/mitogen/pull/715#issuecomment-719266420 migrating to Azure for now
+    # https://github.com/dw/mitogen/pull/715#issuecomment-719266420 migrating to Azure for now due to Travis 50 min time limit cap
+    # azure lets us adjust the cap, and the current STRATEGY=linear tests take up to 1.5 hours to finish
     # - python: "2.7"
     #   env: MODE=ansible VER=2.10.0 DISTROS=debian STRATEGY=linear
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,12 +33,13 @@ script:
 matrix:
   include:
     # Debops tests.
+    # NOTE: debops tests turned off for Ansible 2.10: https://github.com/debops/debops/issues/1521
     # 2.10; 3.6 -> 2.7
-    - python: "3.6"
-      env: MODE=debops_common VER=2.10.0
+    # - python: "3.6"
+    #   env: MODE=debops_common VER=2.10.0
     # 2.10; 2.7 -> 2.7
-    - python: "2.7"
-      env: MODE=debops_common VER=2.10.0
+    # - python: "2.7"
+    #   env: MODE=debops_common VER=2.10.0
 
     # Sanity check against vanilla Ansible. One job suffices.
     - python: "2.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,10 @@ install:
 
 script:
 - .ci/spawn_reverse_shell.py
-- .ci/${MODE}_tests.py
+- MODE=${MODE} .ci/travis.sh
 
+# Travis has a 4MB log limit (https://github.com/travis-ci/travis-ci/issues/1382), but verbose Mitogen logs run larger than that
+# in order to keep verbosity to debug a build failure, will run with this workaround: https://stackoverflow.com/a/26082445
 
 # To avoid matrix explosion, just test against oldest->newest and
 # newest->oldest in various configuartions.
--- a/ansible_mitogen/planner.py
+++ b/ansible_mitogen/planner.py
@@ -59,7 +59,8 @@ import ansible_mitogen.target
 LOG = logging.getLogger(__name__)
 NO_METHOD_MSG = 'Mitogen: no invocation method found for: '
 NO_INTERPRETER_MSG = 'module (%s) is missing interpreter line'
-NO_MODULE_MSG = 'The module %s was not found in configured module paths.'
+# NOTE: Ansible 2.10 no longer has a `.` at the end of NO_MODULE_MSG error
+NO_MODULE_MSG = 'The module %s was not found in configured module paths'
 
 _planner_by_path = {}
 
--- a/ansible_mitogen/planner.py
+++ b/ansible_mitogen/planner.py
@@ -487,7 +487,7 @@ def _propagate_deps(invocation, planner, context):
         # modules=planner.get_module_deps(), TODO
         overridden_sources=invocation._overridden_sources,
         # needs to be a list because can't unpickle() a set()
-        extra_sys_paths=list(invocation._extra_sys_paths)
+        extra_sys_paths=list(invocation._extra_sys_paths),
     )
 
 
--- a/ansible_mitogen/planner.py
+++ b/ansible_mitogen/planner.py
@@ -553,13 +553,14 @@ def _fix_py35(invocation, module_source):
     We replace a relative import in the setup module with the actual full file path
     This works in vanilla Ansible but not in Mitogen otherwise
     """
-    if invocation.module_name == 'setup' and \
+    if invocation.module_name in {'ansible.builtin.setup', 'setup'} and \
             invocation.module_path not in invocation._overridden_sources:
         # in-memory replacement of setup module's relative import
         # would check for just python3.5 and run this then but we don't know the
         # target python at this time yet
+        # NOTE: another ansible 2.10-specific fix: `from ..module_utils` used to be `from ...module_utils`
         module_source = module_source.replace(
-            b"from ...module_utils.basic import AnsibleModule",
+            b"from ..module_utils.basic import AnsibleModule",
             b"from ansible.module_utils.basic import AnsibleModule"
         )
         invocation._overridden_sources[invocation.module_path] = module_source
--- a/ansible_mitogen/planner.py
+++ b/ansible_mitogen/planner.py
@@ -43,8 +43,7 @@ import os
 import random
 
 from ansible.executor import module_common
-from ansible.galaxy.collection import find_existing_collections
-from ansible.utils.collection_loader import AnsibleCollectionConfig
+from ansible.collections.list import list_collection_dirs
 import ansible.errors
 import ansible.module_utils
 import ansible.release
@@ -572,12 +571,8 @@ def _load_collections(invocation):
     Goes through all collection path possibilities and stores paths to installed collections
     Stores them on the current invocation to later be passed to the master service
     """
-    for path in AnsibleCollectionConfig.collection_paths:
-        if os.path.isdir(path):
-            collections = find_existing_collections(path, fallback_metadata=True)
-
-            for collection in collections:
-                invocation._extra_sys_paths.add(collection.b_path.decode('utf-8'))
+    for collection_path in list_collection_dirs():
+        invocation._extra_sys_paths.add(collection_path.decode('utf-8'))
 
 
 def invoke(invocation):
--- a/mitogen/master.py
+++ b/mitogen/master.py
@@ -559,6 +559,9 @@ class SysModulesMethod(FinderMethod):
         """
         Find `fullname` using its :data:`__file__` attribute.
         """
+        # see if collections are being loaded in sys.path here
+        if 'synchronize' in fullname:
+            raise ValueError(sys.path)
         module = sys.modules.get(fullname)
         if not isinstance(module, types.ModuleType):
             LOG.debug('%r: sys.modules[%r] absent or not a regular module',
--- a/mitogen/master.py
+++ b/mitogen/master.py
@@ -522,9 +522,6 @@ class PkgutilMethod(FinderMethod):
 
             # workaround for special python modules that might only exist in memory
             if is_special and is_pkg and not source:
-                # jjj
-                # hope this raises an error
-                raise ValueError(fullname)
                 source = '\n'
         except (AttributeError, ImportError):
             # - Per PEP-302, get_source() and is_package() are optional,
--- a/mitogen/master.py
+++ b/mitogen/master.py
@@ -90,10 +90,10 @@ RLOG = logging.getLogger('mitogen.ctx')
 
 
 # there are some cases where modules are loaded in memory only, such as
-# ansible collections, and the module "filename" is something like __synthetic__
-# which doesn't actually exist
+# ansible collections, and the module "filename" doesn't actually exist
 SPECIAL_FILE_PATHS = {
     "__synthetic__",
+    "<ansible_synthetic_collection_package>"
 }
 
 
@@ -146,7 +146,7 @@ def is_stdlib_path(path):
     )
 
 
-def get_child_modules(path):
+def get_child_modules(path, fullname):
     """
     Return the suffixes of submodules directly neated beneath of the package
     directory at `path`.
@@ -155,12 +155,19 @@ def get_child_modules(path):
         Path to the module's source code on disk, or some PEP-302-recognized
         equivalent. Usually this is the module's ``__file__`` attribute, but
         is specified explicitly to avoid loading the module.
+    :param str fullname:
+        Name of the package we're trying to get child modules for
 
     :return:
         List of submodule name suffixes.
     """
-    it = pkgutil.iter_modules([os.path.dirname(path)])
-    return [to_text(name) for _, name, _ in it]
+    mod_path = os.path.dirname(path)
+    if mod_path != '':
+        return [to_text(name) for _, name, _ in pkgutil.iter_modules([mod_path])]
+    else:
+        # we loaded some weird package in memory, so we'll see if it has a custom loader we can use
+        loader = pkgutil.find_loader(fullname)
+        return [to_text(name) for name, _ in loader.iter_modules(None)] if loader else []
 
 
 def _looks_like_script(path):
@@ -993,7 +1000,7 @@ class ModuleResponder(object):
             self.minify_secs += mitogen.core.now() - t0
 
         if is_pkg:
-            pkg_present = get_child_modules(path)
+            pkg_present = get_child_modules(path, fullname)
             self._log.debug('%s is a package at %s with submodules %r',
                             fullname, path, pkg_present)
         else:
--- a/mitogen/master.py
+++ b/mitogen/master.py
@@ -212,6 +212,10 @@ def _py_filename(path):
     if basepath in SPECIAL_FILE_PATHS:
         return path, True
 
+    # return None, False means that the filename passed to _py_filename does not appear
+    # to be python, and code later will handle when this function returns None
+    # see https://github.com/dw/mitogen/pull/715#discussion_r532380528 for how this
+    # decision was made to handle non-python files in this manner
     return None, False
 
 
--- a/mitogen/master.py
+++ b/mitogen/master.py
@@ -559,9 +559,6 @@ class SysModulesMethod(FinderMethod):
         """
         Find `fullname` using its :data:`__file__` attribute.
         """
-        # see if collections are being loaded in sys.path here
-        if 'synchronize' in fullname:
-            raise ValueError(sys.path)
         module = sys.modules.get(fullname)
         if not isinstance(module, types.ModuleType):
             LOG.debug('%r: sys.modules[%r] absent or not a regular module',
--- a/tests/ansible/integration/action/synchronize.yml
+++ b/tests/ansible/integration/action/synchronize.yml
@@ -44,22 +44,23 @@
     # exception:     self._remove_tmp_path(self._connection._shell.tmpdir)
     # exception: AttributeError: 'get_with_context_result' object has no attribute '_shell'
     # TODO: looks like a bug on Ansible's end with 2.10? Maybe 2.10.1 will fix it
-    - name: do synchronize test
-      block:
-        - synchronize:
-            private_key: /tmp/synchronize-action-key
-            dest: /tmp/sync-test.out
-            src: /tmp/sync-test/
+    # https://github.com/dw/mitogen/issues/746
+    # - name: do synchronize test
+    #   block:
+    #     - synchronize:
+    #         private_key: /tmp/synchronize-action-key
+    #         dest: /tmp/sync-test.out
+    #         src: /tmp/sync-test/
 
-        - slurp:
-            src: /tmp/sync-test.out/item
-          register: out
+    #     - slurp:
+    #         src: /tmp/sync-test.out/item
+    #       register: out
 
-        - set_fact: outout="{{out.content|b64decode}}"
+    #     - set_fact: outout="{{out.content|b64decode}}"
 
-        - assert:
-            that: outout == "item!"
-      when: is_mitogen
+    #     - assert:
+    #         that: outout == "item!"
+    #   when: is_mitogen
 
     # TODO: https://github.com/dw/mitogen/issues/692
     # - file:
--- a/tests/ansible/integration/action/synchronize.yml
+++ b/tests/ansible/integration/action/synchronize.yml
@@ -45,22 +45,22 @@
     # exception: AttributeError: 'get_with_context_result' object has no attribute '_shell'
     # TODO: looks like a bug on Ansible's end with 2.10? Maybe 2.10.1 will fix it
     # https://github.com/dw/mitogen/issues/746
-    # - name: do synchronize test
-    #   block:
-    #     - synchronize:
-    #         private_key: /tmp/synchronize-action-key
-    #         dest: /tmp/sync-test.out
-    #         src: /tmp/sync-test/
+    - name: do synchronize test
+      block:
+        - synchronize:
+            private_key: /tmp/synchronize-action-key
+            dest: /tmp/sync-test.out
+            src: /tmp/sync-test/
 
-    #     - slurp:
-    #         src: /tmp/sync-test.out/item
-    #       register: out
+        - slurp:
+            src: /tmp/sync-test.out/item
+          register: out
 
-    #     - set_fact: outout="{{out.content|b64decode}}"
+        - set_fact: outout="{{out.content|b64decode}}"
 
-    #     - assert:
-    #         that: outout == "item!"
-    #   when: is_mitogen
+        - assert:
+            that: outout == "item!"
+      when: False
 
     # TODO: https://github.com/dw/mitogen/issues/692
     # - file:
--- a/tests/ansible/integration/async/runner_one_job.yml
+++ b/tests/ansible/integration/async/runner_one_job.yml
@@ -40,15 +40,14 @@
       - result1.changed == True
       # ansible/b72e989e1837ccad8dcdc926c43ccbc4d8cdfe44
       - |
-        (ansible_version.full >= '2.8' and
+        (ansible_version.full is version('2.8', ">=") and
          result1.cmd == "echo alldone;\nsleep 1;\n") or
-        (ansible_version.full < '2.8' and
+        (ansible_version.full is version('2.8', '<') and
          result1.cmd == "echo alldone;\n sleep 1;")
       - result1.delta|length == 14
       - result1.start|length == 26
       - result1.finished == 1
       - result1.rc == 0
-      - result1.start|length == 26
 
   - assert:
       that:
@@ -56,10 +55,9 @@
       - result1.stderr_lines == []
       - result1.stdout == "alldone"
       - result1.stdout_lines == ["alldone"]
-    when: ansible_version.full > '2.8'  # ansible#51393
+    when: ansible_version.full is version('2.8', '>')  # ansible#51393
 
   - assert:
       that:
       - result1.failed == False
-    when: ansible_version.full > '2.4'
-
+    when: ansible_version.full is version('2.4', '>')
--- a/tests/ansible/integration/connection_loader/paramiko_unblemished.yml
+++ b/tests/ansible/integration/connection_loader/paramiko_unblemished.yml
@@ -1,12 +1,13 @@
 # Ensure paramiko connections aren't grabbed.
 
-- name: integration/connection_loader/paramiko_unblemished.yml
-  hosts: test-targets
-  any_errors_fatal: true
-  tasks:
-    - custom_python_detect_environment:
-      connection: paramiko
-      register: out
+# TODO: this is flaky -> https://github.com/dw/mitogen/issues/747
+# - name: integration/connection_loader/paramiko_unblemished.yml
+#   hosts: test-targets
+#   any_errors_fatal: true
+#   tasks:
+#     - custom_python_detect_environment:
+#       connection: paramiko
+#       register: out
 
-    - assert:
-        that: not out.mitogen_loaded
+#     - assert:
+#         that: not out.mitogen_loaded
--- a/tests/ansible/integration/connection_loader/paramiko_unblemished.yml
+++ b/tests/ansible/integration/connection_loader/paramiko_unblemished.yml
@@ -2,13 +2,15 @@
 ---
 
 # TODO: this is flaky -> https://github.com/dw/mitogen/issues/747
-# - name: integration/connection_loader/paramiko_unblemished.yml
-#   hosts: test-targets
-#   any_errors_fatal: true
-#   tasks:
-#     - custom_python_detect_environment:
-#       connection: paramiko
-#       register: out
+- name: integration/connection_loader/paramiko_unblemished.yml
+  hosts: test-targets
+  any_errors_fatal: true
+  tasks:
+    - debug:
+        msg: "skipped for now"
+    # - custom_python_detect_environment:
+    #   connection: paramiko
+    #   register: out
 
-#     - assert:
-#         that: not out.mitogen_loaded
+    # - assert:
+    #     that: not out.mitogen_loaded
--- a/tests/ansible/integration/connection_loader/paramiko_unblemished.yml
+++ b/tests/ansible/integration/connection_loader/paramiko_unblemished.yml
@@ -1,4 +1,5 @@
 # Ensure paramiko connections aren't grabbed.
+---
 
 # TODO: this is flaky -> https://github.com/dw/mitogen/issues/747
 # - name: integration/connection_loader/paramiko_unblemished.yml
--- a/tests/ansible/integration/connection_loader/paramiko_unblemished.yml
+++ b/tests/ansible/integration/connection_loader/paramiko_unblemished.yml
@@ -1,16 +1,18 @@
 # Ensure paramiko connections aren't grabbed.
 ---
 
-# TODO: this is flaky -> https://github.com/dw/mitogen/issues/747
 - name: integration/connection_loader/paramiko_unblemished.yml
   hosts: test-targets
   any_errors_fatal: true
   tasks:
     - debug:
         msg: "skipped for now"
-    # - custom_python_detect_environment:
-    #   connection: paramiko
-    #   register: out
+    - name: this is flaky -> https://github.com/dw/mitogen/issues/747
+      block:
+        - custom_python_detect_environment:
+          connection: paramiko
+          register: out
 
-    # - assert:
-    #     that: not out.mitogen_loaded
+        - assert:
+            that: not out.mitogen_loaded
+      when: False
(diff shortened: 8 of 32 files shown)


cognifloyd:
Hmm. Fascinating.
the mixin is using `ansible.plugins.action.ActionBase` but the error message I saw said it could not find `ansible_collections.ansible.builtin.plugins.action`. So, we're hitting something where ansible is prepending `ansible_collections.` and then trying to look it up.
```
    class ActionModuleMixin(ansible.plugins.action.ActionBase):
AttributeError: module 'ansible_collections.ansible.builtin.plugins.action' has no attribute 'ActionBase'
```

In https://github.com/ansible/ansible/pull/52194 there was a line that prefixed modules with `ansible_collections.`, but I don't see it in the stable-2.9 branch.

[This is a concerning comment](https://github.com/ansible/ansible/blob/28e9121d94e66e46d72bfbf3c9db0802417c5a16/lib/ansible/utils/collection_loader.py#L132-L136) "preventing controller-side code injection during collection loading"... I wonder if that's what's breaking this.
```python
        # this loader implements key functionality for Ansible collections
        # * implicit distributed namespace packages for the root Ansible namespace (no pkgutil.extend_path hackery reqd)
        # * implicit package support for Python 2.7 (no need for __init__.py in collections, except to use standard Py2.7 tooling)
        # * preventing controller-side code injection during collection loading
        # * (default loader would execute arbitrary package code from all __init__.py's)
```

I'll keep looking for how `ansible_collections.` is getting prepended.

s1113950:
Here's where it gets prepended:
https://github.com/ansible/ansible/pull/52194/files#diff-279e661b06196c21d04f8226e65ee458R24

They alias to ansible's modules and build `ansible_collections.*` there

s1113950:
so https://github.com/ansible/ansible/pull/52194/files#diff-279e661b06196c21d04f8226e65ee458R27 specifically is why we see the error about `ansible_collections.ansible.builtin.plugins.action`, because that's pointing to `ansible.plugins.action`, but if you check out the `collection_loader` it doesn't have an `actionbase`. They put the collection loader stuff in a util file and use it as a special importer path rather than a specific loader type: https://github.com/ansible/ansible/pull/52194/files#diff-cb51550dcfdbd27a3f224be2de15e2e8R822

s1113950:
Ok wat. I'm able to dynamically load things now from sys.modules that are collections, but in this example: https://github.com/alikins/collection_inspect/tree/master/plugins only action and modules appear in sys.modules:
```
(Epdb) dir(sys.modules["ansible_collections.alikins.collection_inspect.plugins"])
['__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'action', 'modules']
```
so I'm gonna have to go the full file path route now possibly, or I'll try one of the other *method classes used to detect source code! :D challenge_accepted

cognifloyd:
So, that collection has:
- `action` :heavy_check_mark:
- `callback` :question: _this might require whitelisting in ansible.cfg?_
- `filter` :question:
- `lookup` :question:
- `module_utils` :question: **comments below**
- `modules` :heavy_check_mark: _this uses a special `AnsibleFlatMapLoader`_
- `vars` :x: _I don't expect this one to work until 2.10_

Both action and module_utils are special cases in `_SYNTHETIC_PACKAGES`. Both are maps, but module_utils has `graft=True`, so maybe that is the difference?
graft is used in `AnsibleCollectionLoader.__init__` to "pre-inject grafted package maps" and again in `AnsibleCollectionLoader._find_module`. In there, it should be doing `import_module("ansible.module_utils" + ".whatever")`.

Ah. I suspect the only way `module_utils` would get loaded on the controller side is if one of the other plugins in the collection imports something from it. AnsiballZ (`lib/ansible/executor/module_common.py`) takes care of detecting module_utils to shove down to the store.

Here's a significant change that adjusted how AnsiballZ collects files to be able to load module_utils with relative imports from collections into the AnsiballZ wrapper: https://github.com/ansible/ansible/pull/61196

cognifloyd:
Oh, and there's a  `CollectionModuleInfo.get_source()` method there too. You had a comment about getting the source, maybe using that will help.

s1113950:
Thanks for the info 👍

nitzmahone:
FYI: you might not want to go too crazy here just yet- 2.10 changes the internals on this stuff quite a bit for 2.9 backwards compatibility with collectionized content that's no longer in Ansible core. See https://github.com/ansible/ansible/pull/67684/ which is just about to land in devel...

s1113950:
Thanks for the heads up!

cognifloyd:
Does that mean that mitogen should only start supporting collections with 2.10?
I suspect that ansible's moving modules/plugins from ansible core to collections in 2.10 might require (many?) other changes in ansible_mitogen as well (and maybe mitogen itself too). The 2.10 move might be a forcing function: without collections support, ansible_mitogen will not be very useful.

Plus, with mitogen's philosophy of supporting very old python, and very old ansible versions, what does this mean for ansible_mitogen supporting ansible 2.8 (w/ tech preview of collections) or 2.9 (w/ support for collections)?

nitzmahone:
@cognifloyd Not sure if you're asking me or others on this PR. I've got no opinion- someone on our team pointed this PR out to me since I'm doing most of the work on the core engine's collections support (which is the part you're most likely to be impacted by). I will say that 2.10 will be the first release where *everyone* gets exposure to collections. Only the bleeding-edge folks and partners developing collections were doing much with them in 2.8/2.9, but the 2.10 community distributions of Ansible will all be collection-ized internally (ansible-base + all the collections required to maintain 2.9 compatibility "installed" to `sys.path`, likely).

cognifloyd:
@nitzmahone I was talking more about the path forward for mitogen. Given that ansible mitogen advertises support for ansible 2.3 to 2.8 (see [here](https://mitogen.networkgenomics.com/ansible_detailed.html#noteworthy-differences)), do we need to worry about not supporting collections with ansible 2.8 or 2.9? At this point, the answer is probably "no" we don't need to support collections there. We'll need to document that and get the website updated (is that in this repo? if it is, I can submit a PR).

I understand that there are a lot of changes with the 2.10 release. I'd really like to try out mitogen with collections.
So, @s1113950, what if we adjusted the tests temporarily to pull in ansible devel branch to start figuring out what is needed with all the new collections stuff. That way we can release the next version of mitogen with 2.10 support soon after 2.10 is released.

s1113950:
Website gets updated in this way in the code: https://github.com/dw/mitogen/pull/714 

Using `devel` makes sense to me. If a ton of stuff is gonna change it doesn't make much sense to me to spend a ton of time now getting things to work on 2.9 just for it to change a bunch in 2.10. I'll adjust tests on this PR to point to the latest hash of `devel` and go from there!

I don't know what to do about the backwards compat issue @cognifloyd. We already have quite a few places in the code that do hackish things to support super old Ansible... I wonder if it makes sense to only support ansible 2.10+ with mitogen 0.3.0+ and tell folks to use mitogen < 0.3 if they want <= ansible 2.9, depending on the kinds of changes that 2.10 brings? We could do a 0.2.10 release with everything leading up to 2.10 support (like ansible_python_interpreter detection) and then do a version change to 0.3 for ansible 2.10+ perhaps 🤔 WDYT?

cognifloyd:
I guess we need @dw to weigh in and say whether this would work for him and how he's using it in Operon. From everything I've read, backwards compat is super important for him.

s1113950:
Agreed 👍

s1113950:
@cognifloyd David is cool with bumping the version number for collections support :)

cognifloyd:
By bumping the version number, you mean bumping the minimum ansible version, right?
Sounds awesome! All in on 2.10!

Leletir:
Hi, 

First of all thank you very much for your work ! Do you have any idea when you will release the collection support ?

s1113950:
@Leletir the current plan is sometime after Ansible 2.10 lands. I ran into a blocker for a long time (with #723) + some personal stuff happened which prevented me from working on this. Additionally, there have been some small Mitogen tickets that have come in that I've taken care of as well.
My plan is to finish the last small ticket I'm on and then finish this up. I was putting this off partially because of the test failures and partially because Ansible 2.10 support was months away at the time so I wanted to wait until bleeding-edge Ansible was more stable before continuing.

cognifloyd:
2.10 is fresh off the press, released this week: https://groups.google.com/g/ansible-announce/c/RC6ysUy961o
So, maybe we can address this soonish.

s1113950:
:scream: I tried `pip install ansible` today but didn't try the new `ansible-base`. I'll reprioritize this then.

s1113950:
https://github.com/ansible/ansible/blob/v2.9.12/lib/ansible/plugins/connection/ssh.py#L558 is what the sshpass check used to be, which worked for Mitogen

s1113950:
🎉 @cognifloyd @Leletir  🎉 PR isn't done yet but I was able to run a `docker_container` task installed via `ansible-galaxy collection install community.general`. Please feel free to test it and see if I missed anything 😄 . Also see the updated PR description for a description of what's changed and further TODOs.

Leletir:
Hello guys,

Thank you very much for your work! I will try it when I come back to work.
Have a good day :)

mitogen/master.py
```diff
@@ -189,6 +197,11 @@ def _py_filename(path):
     if os.path.exists(path) and _looks_like_script(path):
         return path
 
+    basepath = os.path.basename(path)
+    for filename in SPECIAL_FILE_PATHS:
+        if basepath == filename:
+            return path
```
cognifloyd, line 203: I think this could be simplified. Is your version more performant somehow?
```suggestion
    if basepath in SPECIAL_FILE_PATHS:
        return path
```

s1113950: :man_facepalming: you're correct! I'll make the SPECIAL_FILE_PATHS a set and compare against basepath as you suggest :sweat_smile: good catch

s1113950:
Am now able to run the `alikins` example from https://github.com/dw/mitogen/issues/652#issuecomment-602107590

s1113950:
Lots of tests are passing after I set it to only test Ansible 2.10! :D

s1113950:
`debops` tests are failing because of
```
TASK [pki : DebOps pre_tasks hook] *********************************************
fatal: [debops-target-debian-1]: FAILED! => {"msg": "lookup plugin (task_src) not found"}
fatal: [debops-target-debian-2]: FAILED! => {"msg": "lookup plugin (task_src) not found"}
```
which lives here: https://github.com/debops/debops/blob/master/ansible/roles/ansible_plugins/lookup_plugins/task_src.py

cognifloyd:
I haven't used DebOps. It looks like that's not a proper collection, so how are those plugins getting added to the [lookup plugin path](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-lookup-plugin-path)?

I don't think collections support including plugins under roles.

What happens if you drop the mitogen injection here? If it doesn't work, maybe DebOps is simply incompatible with ansible 2.10?
https://github.com/dw/mitogen/blob/8d95172821410dd93dd0de4c5e2bc95ebe54a314/.ci/debops_common_tests.py#L29-L35

s1113950:
> I haven't used DebOps. It looks like that's not a proper collection, so how are those plugins getting added to the [lookup plugin path](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-lookup-plugin-path)?
> 
> I don't think collections support including plugins under roles.
> 
> What happens if you drop the mitogen injection here? If it doesn't work, maybe DebOps is simply incompatible with ansible 2.10?
> https://github.com/dw/mitogen/blob/8d95172821410dd93dd0de4c5e2bc95ebe54a314/.ci/debops_common_tests.py#L29-L35

I'll give it a try! 😄 I already found 1 issue that won't work with Ansible 2.10: https://github.com/debops/debops/issues/723

s1113950:
@cognifloyd I cannot figure out why `synchronize` isn't loading properly in Azure 😞 it works fine for me locally running either python2.7 or python3. Collections seem to be loading because otherwise other stuff would fail earlier but `ansible.posix.synchronize` isn't being called (in the PR I deliberately tried to error out if `synchronize` was detected in Mitogen but it never errored; it errors as expected locally though). I see the call to `redirecting...to ansible.posix.synchronize` in the error logs which means Ansible is detecting that the `ansible.posix.synchronize` collection is installed, but then it fails with a traceback that shouldn't be called because Mitogen has its own `remove_tmp_path` function that's called instead.

s1113950:
I made https://github.com/dw/mitogen/issues/746 ; I can't spend any more time on getting synchronize to work in Azure now

jhf2442:
Any ETA for the release of a release that's ansible 2.10 compatible ?
```
ERROR! Your Ansible version (2.10.1) is too recent. The most recent version
supported by Mitogen for Ansible is (2, 9).x. Please check the Mitogen
```

s1113950:
@jhf2442 I'm hoping in the next week or two there'll be a release cut, and then most likely a minor version bump after that to fix any major regressions found (there's a ton of stuff that's changed from Ansible < 2.10 so it's very likely some edge cases will be missed). Depending on how big of a change Mitogen needs there'll also be a separate version cut that includes `master` leading up to this PR, and not include 2.10 support. See PR comment history if you're curious; TL;DR it's becoming a lot of effort to support super old versions of Python/Ansible as Ansible keeps changing its structure around, so depending on what needs changing, the new Mitogen might only support 2.10+ (esp. if things like performance start to take a hit).

Biggest issue now is I'm currently trying to make sure all test cases pass. Lately it's been slow-going because much of Mitogen's test framework is old, and what works locally for my on my laptop wasn't working in Azure, for example (so that's why there are so many commits; gotta commit and then check logs and see what's up. There is also a bug in Azure Devops where if your log is long enough when you try and scroll it at the very bottom it skips, which makes seeing the actual failure harder 😞 ).

You're welcome to test this branch out and post any findings here! Would definitely appreciate the debugging help.

s1113950:
https://github.com/dw/mitogen/issues/747 paramiko fails in some runs and not in others, so making a ticket and continuing on for now 😕

s1113950:
Working on the latest test failure now: mixing mitogen + regular Ansible `strategy` plugins. Added some updated debugging info to the PR description

s1113950:
Will try to rerun Azure tests later. 
Run from yesterday had no problems creating users: https://dev.azure.com/dw-mitogen/Mitogen/_build/results?buildId=1056&view=logs&j=c33a2a27-ad1b-5f95-a372-4c6c65381ab9
Run from today has error codes that indicate there's something wrong with admin perms on the Mac host: https://dev.azure.com/dw-mitogen/Mitogen/_build/results?buildId=1060&view=logs&j=c33a2a27-ad1b-5f95-a372-4c6c65381ab9 
```
 "msg": "Error when changing password", 
 "out": "passwd: DS error: eDSAuthFailed\n", 

"msg": "Cannot create user \"mitogen__user5\".", 
 "out": "<dscl_cmd> DS Error: -14009 (eDSUnknownNodeName)\n", 
```
 🤦 🤦 🤦  I'm hoping it's a temp Azure thing. Azure tests should all be passing at this point based on the diff between the two test runs.

Something definitely changed with Azure because this used to work:
```
Running: ['/usr/bin/time', '--', 'brew', 'install', 'http://git.io/sshpass.rb']
Error: Calling Non-checksummed download of sshpass formula file from an arbitrary URL is disabled! Use 'brew extract' or 'brew create' and 'brew tap-new' to create a formula file in a tap on GitHub instead.
```
I installed sshpass 1.06 to see if that'd fix it (the git.io url pointed to 1.05), that still failed so I manually installed 1.05 without brew but also still seeing fails... 😞

s1113950:
🎉 Azure tests pass now 🎉

s1113950:
Going to disable DebOps tests for now. After beating my head against the wall for awhile I found: https://github.com/debops/debops/issues/1521 which I'm running into both with and without Mitogen. Here's what I attempted: https://github.com/s1113950/mitogen-test/blob/master/roles/run_test/debops/tests.yml

s1113950:
😞 native Ansible 2.10 tests are taking > 50 min. That's a problem because that's how long Travis free tests are allowed to run for. I'll split them apart, weird tho that it times out so it might be some other issue.

ivandeex:
> native Ansible 2.10 tests are taking > 50 min.
> that's how long Travis free tests are allowed to run for.

Free Travis is very slow these days.
They are migrating travis-ci.org to travis-ci.com now
Take a look at https://docs.travis-ci.com/user/migrate/open-source-repository-migration#migrating-a-repository
Hopefully it fixes slow tests for you

s1113950:
@ivandeex thanks for the info! Unfortunately I can't do this migration since I don't have access to the `Settings` of Mitogen. I read over your link and this part won't work for me: https://docs.travis-ci.com/user/migrate/open-source-repository-migration#do-i-need-to-make-any-other-changes 
```
“Protected Branches” feature in GitHub to require a passing Travis CI build before merging a Pull Request: make sure to edit your required status checks to now use Travis CI - Pull Request and/or Travis CI - Branch. See: Required status checks at GitHub.
```
Travis is `required` on Mitogen's Github and I can't change it to use the new check.

@dw can you do this migration? In the meantime I'll try and either split out the vanilla Ansible run or put it on Azure (I've noticed Azure's tests also start up a lot faster than Travis's)

s1113950:
Moving it to Azure didn't work, a vanilla Ansible 2.10 test is hanging but works fine locally 🤔 I wonder if it's a Mac vs Debian thing. Will try moving them to running on Mac instead.

s1113950:
🎉 At long last... 🎉 
One of the longer tests has Mitogen taking 16 min 13 sec, while vanilla Ansible took 1 hr 17 min 32 sec.
This is despite lots of tests being ignored unless running with Mitogen :D (this is mostly due to `filetop` tests taking forever under vanilla Ansible)

Anyone who wants to test this patch, it would be much appreciated. I tried to find every normal use case I could find to test with, but there are so many uses for Ansible that something might have been missed.

@cognifloyd can you do an initial review here? 🙏  I tried to note places in the code that would be backwards-incompatible. I'm still unsure if we want to have a ton of `if` statements everywhere to check for Ansible version and change logic as such to support super old versions, given how much Ansible's API changed. Additionally, I turned off tests for older Ansible versions, but if we want to keep backwards-compat then I'll need to add those back in. I'll change the README based on what we decide.

I wonder if this should be an alpha release and then we can fix any regressions that have been missed (there are also a few new Github tickets that would be good to get in as well)? Or should we just go for it and do a `0.3.1` release later on?

extmind:
I've tested the current version with some playbooks and mostly it worked fine. However, when connecting to a host running Debian stretch (oldstable) the setup module fails executing with the following error:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: SystemError: Parent module 'ansible.modules' not loaded, cannot perform relative import
host | FAILED! => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python3"
    },
    "changed": false,
    "module_stderr": "Traceback (most recent call last):\n  File \"master:/.../mitogen/ansible_mitogen/runner.py\", line 975, in _run\n    self._run_code(code, mod)\n  File \"master:/.../mitogen/ansible_mitogen/runner.py\", line 939, in _run_code\n    exec(code, vars(mod))\n  File \"master:/usr/lib/python3.8/site-packages/ansible/modules/setup.py\", line 130, in <module>\nSystemError: Parent module 'ansible.modules' not loaded, cannot perform relative import\n",
    "module_stdout": "",
    "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error",
    "rc": 1
}
```
Also, on execution of _some_ modules, warnings are emitted repeatedly for about every python module installed on the system:
```
[WARNING]: Collection at '/usr/lib/python3.8/site-packages/meld/ui' does not have a galaxy.yml or a MANIFEST.json file, cannot detect version.
[WARNING]: Collection at '/usr/lib/python3.8/site-packages/meld/matchers' does not have a galaxy.yml or a MANIFEST.json file, cannot detect version.
[WARNING]: Collection at '/usr/lib/python3.8/site-packages/meld/vc' does not have a galaxy.yml or a MANIFEST.json file, cannot detect version.
[WARNING]: Collection at '/usr/lib/python3.8/site-packages/meld/__pycache__' does not have a galaxy.yml or a MANIFEST.json file, cannot detect version.
[...]
```

s1113950:
Thanks for testing! 😄 I'll see about disabling those warnings and patching debian 9.

s1113950:
@extmind I fixed the warning issue, and improved how collection paths load in the process! Thanks again for pointing that out.

I tried replicating running Mitogen on Debian 9 but it worked for me 🤔 I tried a `debian:9` docker container and ran `apt-get update && apt-get install -y python3` on it to give it Python. Can you describe your debian setup?

.ci/localhost_ansible_tests.py
```diff
@@ -20,12 +20,15 @@
 
 
 with ci_lib.Fold('job_setup'):
-    # Don't set -U as that will upgrade Paramiko to a non-2.6 compatible version.
-    run("pip install -q virtualenv ansible==%s", ci_lib.ANSIBLE_VERSION)
```
moreati, line 24: Where did the installation of ansible move to? I don't see a corresponding change in `localhost_ansible_install.py`.

ansible_mitogen/planner.py
```diff
@@ -478,8 +484,10 @@ def _propagate_deps(invocation, planner, context):
 
         context=context,
         paths=planner.get_push_files(),
-        modules=planner.get_module_deps(),
-        overridden_sources=invocation._overridden_sources
+        # modules=planner.get_module_deps(), TODO
+        overridden_sources=invocation._overridden_sources,
+        # needs to be a list because can't unpickle() a set()
+        extra_sys_paths=list(invocation._extra_sys_paths)
```
moreati, line 490: Bikeshed: Suggest using a trailing  comma, to make future diffs cleaner.

mitogen/master.py
```diff
@@ -177,17 +192,27 @@ def _looks_like_script(path):
...
 
     if os.path.exists(path) and _looks_like_script(path):
-        return path
+        return path, False
+
+    basepath = os.path.basename(path)
+    if basepath in SPECIAL_FILE_PATHS:
+        return path, True
+
+    return None, False
```
moreati, line 215: Is there a non-error situation that would cause the function to reach this point? In other words, is there a danger that this line would mask a mistake elsewhere in the code? If so, would it make more sense to raise an exception, e.g. ValueError?

tests/ansible/integration/action/synchronize.yml
```diff
@@ -40,19 +40,27 @@
...
+    # exception:     self._remove_tmp_path(self._connection._shell.tmpdir)
+    # exception: AttributeError: 'get_with_context_result' object has no attribute '_shell'
+    # TODO: looks like a bug on Ansible's end with 2.10? Maybe 2.10.1 will fix it
+    # https://github.com/dw/mitogen/issues/746
+    # - name: do synchronize test
+    #   block:
+    #     - synchronize:
+    #         private_key: /tmp/synchronize-action-key
+    #         dest: /tmp/sync-test.out
+    #         src: /tmp/sync-test/
```
moreati, line 53: Suggestion: Skip with a `when` clause, rather than commenting them out. That way they're still visible in test output, and so less likely to be forgotten.

tests/ansible/integration/async/runner_one_job.yml
```diff
@@ -40,26 +40,24 @@
       - result1.changed == True
       # ansible/b72e989e1837ccad8dcdc926c43ccbc4d8cdfe44
       - |
-        (ansible_version.full >= '2.8' and
+        (ansible_version.full is version('2.8', ">=") and
```
moreati, line 43: Rant: Ugh, `thing is version()` is such an ugly wart of Ansible. Not a comment against your code, you've done the right thing. I'm just venting.

tests/ansible/integration/connection_loader/paramiko_unblemished.yml
```diff
@@ -1,12 +1,16 @@
...
   any_errors_fatal: true
   tasks:
-    - custom_python_detect_environment:
-      connection: paramiko
-      register: out
+    - debug:
+        msg: "skipped for now"
+    # - custom_python_detect_environment:
+    #   connection: paramiko
+    #   register: out
```
moreati, line 13: Suggestion: Use a when clause, to make the skipped task more explicit and visible.

tests/ansible/integration/transport_config/become_pass.yml
```diff
@@ -124,7 +125,7 @@
       - out.result|length == 2
       - out.result[0].method == "ssh"
       - out.result[1].method == "sudo"
-      - out.result[1].kwargs.password == "a.b.c"
+      - out.result[1].kwargs.password == "c.b.a"
```
moreati, line 128: I don't understand. What caused this value to be reversed?

tests/requirements.txt
```diff
@@ -3,7 +3,7 @@ coverage==4.5.1
 Django==1.6.11  # Last version supporting 2.6.
 mock==2.0.0
 pytz==2018.5
-cffi==1.11.2  # Random pin to try and fix pyparser==2.18 not having effect
+cffi==1.14.3  # Random pin to try and fix pyparser==2.18 not having effect
```
moreati, line 6: Have you upgraded cffi on general principal, or was it for something specific?

moreati:
I've pushed tox.ini updates to https://github.com/moreati/mitogen/tree/collectionsSupport, so I can run the tests locally and better understand the testing situation.

moreati:
Pushed 2 more test suite improvements to https://github.com/moreati/mitogen/tree/collectionsSupport. With this revision I see all Tox environments passing

```
  init: commands succeeded
  py27-mitogen: commands succeeded
  py36-mitogen: commands succeeded
  py37-mitogen: commands succeeded
  py27-ansible2.10: commands succeeded
  py36-ansible2.10: commands succeeded
  py37-ansible2.10: commands succeeded
  report: commands succeeded
  congratulations :)
```

s1113950:
> > I tried replicating running Mitogen on Debian 9 but it worked for me thinking I tried a `debian:9` docker container and ran `apt-get update && apt-get install -y python3` on it to give it Python. Can you describe your debian setup?
> 
> These are just some VMs serving as web or mail servers and incidentally have both python 2 and 3 installed.
> I've noticed that when using this PR, ansible detects and chooses python3 but fails. If I set the ansible_python_interpreter variable to python 2, it works. When using a standard ansible strategy python 2 is detected, but both the auto detected python 2 and a manually set python 3 work.

@extmind does running vanilla Ansible work with that debian setup? Mitogen calls `discover_interpreter()` from `Ansible`, so am curious if it works with vanilla Ansible 🤔

.ci/localhost_ansible_tests.py
s1113950: It was already there: https://github.com/dw/mitogen/pull/715/files#diff-ba88d108dcb714bf83179965a30e86a499f9b9e24b87d21592d40387f8394d76L14 😅 I deleted a duplicate install line

mitogen/master.py
s1113950: 🤔 I don't think so. The goal of the function is just to check if a file path is a python file path or not, so it's ok if it returns `False`. There's logic after:
```
if not path:
            return
```
in some places the function is used, which handle what happens if the file isn't Python :)

tests/ansible/integration/action/synchronize.yml
s1113950: 👍 good point!

tests/ansible/integration/transport_config/become_pass.yml
s1113950: Not sure, I tried to search through the Ansible changelog, but it appears to be how password flag handling changed between versions. This is the hosts file for `tc-become-pass-both`: `tc-become-pass-both ansible_become_password=a.b.c ansible_become_pass=c.b.a`, so it means now that `ansible_become_pass` has higher precedence than `ansible_become_password`

tests/requirements.txt
s1113950: I don't remember exactly, but I believe it's because of the new Ansible version. There was a reason, but I don't remember now 🤦 it's a newer version that works though haha, so improvement? 😅

s1113950:
@moreati thanks so much for the review! I made changes you requested and commented on some things. Let me know what you think :)

s1113950:
also @moreati your commits are great 🎉 Mind making a PR for it? :D Would love to get them into the release

extmind:
> > > I tried replicating running Mitogen on Debian 9 but it worked for me thinking I tried a `debian:9` docker container and ran `apt-get update && apt-get install -y python3` on it to give it Python. Can you describe your debian setup?
> > 
> > 
> > These are just some VMs serving as web or mail servers and incidentally have both python 2 and 3 installed.
> > I've noticed that when using this PR, ansible detects and chooses python3 but fails. If I set the ansible_python_interpreter variable to python 2, it works. When using a standard ansible strategy python 2 is detected, but both the auto detected python 2 and a manually set python 3 work.
> 
> @extmind does running vanilla Ansible work with that debian setup? Mitogen calls `discover_interpreter()` from `Ansible`, so am curious if it works with vanilla Ansible thinking

Yes, that's what I tried to tell with my last sentence. Using vanilla Ansible works, discovers Python 2 instead, but will work with both Python versions anyway.

cnkk:
> @extmind I fixed the warning issue, and improved how collection paths load in the process! Thanks again for pointing that out.
> 
> I tried replicating running Mitogen on Debian 9 but it worked for me  I tried a `debian:9` docker container and ran `apt-get update && apt-get install -y python3` on it to give it Python. Can you describe your debian setup?

Hi! This does not work for me on debian stretch.

I have a Debian stretch server running with _Python 3.5.3_
There is also no _Python 2_ installed on the system.

My client:
```shell
ansible 2.10.3
  config file = /home/user/workdir/deployment/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/.local/lib/python3.8/site-packages/ansible
  executable location = /home/user/.local/bin/ansible
  python version = 3.8.5 (default, Jul 28 2020, 12:59:40) [GCC 9.3.0]
```
My ansible command:
```shell
ANSIBLE_STRATEGY_PLUGINS=/home/user/mitogen/mitogen/ansible_mitogen/plugins/strategy ANSIBLE_STRATEGY=mitogen_linear /home/user/.local/bin/ansible serversetup-test01.dev -m ansible.builtin.setup -a "filter=ansible_local" -i inventory -vvv
```
The error output of ansible
```shell
The full traceback is:
  File "master:/home/user/mitogen/mitogen/ansible_mitogen/runner.py", line 975, in _run
    self._run_code(code, mod)
  File "master:/home/user/mitogen/mitogen/ansible_mitogen/runner.py", line 939, in _run_code
    exec(code, vars(mod))
  File "master:/home/user/.local/lib/python3.8/site-packages/ansible/modules/setup.py", line 130, in <module>
SystemError: Parent module 'ansible.modules' not loaded, cannot perform relative import
serversetup-test01.dev | FAILED! => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python3"
    },
    "changed": false,
    "module_stderr": "Traceback (most recent call last):\n  File \"master:/home/user/mitogen/mitogen/ansible_mitogen/runner.py\", line 975, in _run\n    self._run_code(cod
e, mod)\n  File \"master:/home/user/mitogen/mitogen/ansible_mitogen/runner.py\", line 939, in _run_code\n    exec(code, vars(mod))\n  File \"master:/home/user/.local/li
b/python3.8/site-packages/ansible/modules/setup.py\", line 130, in <module>\nSystemError: Parent module 'ansible.modules' not loaded, cannot perform relative import\n",
    "module_stdout": "",
    "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error",
    "rc": 1
}
```

s1113950:
@cnkk @extmind fixed the import error, please give it a try 😄 . The bug is with the `setup` module in Ansible, but specifically only for Pythons 3.5.1-3.5.3 (I tested a bunch of Python3 versions the first time Mitogen had this error). Ansible changed the signature of the `setup` module in 2.10, which broke the hack.

cnkk:
Works now. Thanks for your efforts!

extmind:
> @cnkk @extmind fixed the import error, please give it a try smile . The bug is with the `setup` module in Ansible, but specifically only for Pythons 3.5.1-3.5.3 (I tested a bunch of Python3 versions the first time Mitogen had this error). Ansible changed the signature of the `setup` module in 2.10, which broke the hack.

Thanks for tracking this down. Small addition: When executing a playbook, ansible reported a slightly different error that I was able to circumvent by adding `ansible.legacy.setup` to the set of broken setup modules.

s1113950:
@moreati @dw are you cool with the updates? If so, I'd like to finalize the PR, add some documentation, and tag a release. Big finalization question is based on the changes required, does it make sense to break backwards compat. I've `NOTE:`-ed every place I could find in the PR of some type of api-like breaking change (ie ansible renaming `get` to `get_with_context`). 

I vote doing 2 releases: 1 with everything in `master` except for this PR, and then another with this PR in it. That way if someone wants to use an older version of Ansible then they could use the older Mitogen version, and we wouldn't have to sacrifice speed and code readability by maintaining older Ansible API changes hanging around.

.ci/localhost_ansible_tests.py
moreati: ok

mitogen/master.py
moreati: My concern was that if we *expect* a line of code never to be reached (e.g. the last line of this function), then we should either a) leave it out, or b) write something that fails loudly (e.g. `raise SomeException('Unexpected fall though')`).

We shouldn't silently return a best guess/plausible value.

So, what is the expected condition/circumstance that the final `return None, False` is meant to handle? A code comment to clear that up would make this function easier to understand.

tests/ansible/integration/transport_config/become_pass.yml
moreati: I'm still concerned by this. I'll try to understand what's happened

tests/requirements.txt
moreati: ok

moreati (changes requested):

moreati:
> @moreati @dw are you cool with the updates?

Mostly. I've updated my review comments and I think two still need addressing. Additionally I see there are 5 remaining subtasks

> I vote doing 2 releases: 1 with everything in master

I see the following blockers to releasing current master
1. Travis tests are currently red
2. I don't know the status of Ansible 2.9 support in current master

mitogen/master.py
s1113950: That's a good concern. I double-checked out the places that `_py_filename` is called, and it's all in `mitogen/master.py`, in the various `find` methods of the various ways Mitogen attempts to determine the source code of a module here:
```
get_module_methods = [
        DefectivePython3xMainMethod(),
        PkgutilMethod(),
        SysModulesMethod(),
        ParentEnumerationMethod(),
    ]
```
`_py_filename` is used to attempt to determine the file path of the name of whatever module Mitogen is attempting to load. I think perhaps the function's name could be a bit more descriptive, ie something like `_get_python_file_path`, but the function can also be used as a utility as well, so that might be why the name wasn't changed? I don't really have a strong preference on the name change myself. 

In every case that `_py_filename` is called, handling of an empty `path` result reminds me of how `go` does it, where the result of `_py_filename()` is checked in an `if` statement, and if it's none then the `find()` method returns None and then the next `find()` method is called. I see 2 options here:

1. Add a code comment to explain that `return None, False` means that the filename passed to `_py_filename` does not appear to be python, and code later will handle when this function returns `None`.
2. Change the end of the function to something that fails loudly, then in all places where `_py_filename` is called, handle the failed exception and then make the `find()` functions return at that point if the loud exception is called.

Comparing the 2 options, checking a return type is less expensive than throwing an exception and then handling that exception, so for that reason I'm inclined to go with option 1 given that these `find` functions are called for every python lib that Mitogen tries to load. WDYT?

tests/ansible/integration/transport_config/become_pass.yml
s1113950: My best guess is the change here: https://github.com/ansible/ansible/blame/v2.10.0/lib/ansible/executor/task_executor.py#L1029-L1039 code was changed 6 and 7 months ago around handling `become_pass`, but I'm not positive. I don't really understand it either, but something changed so that `ansible_become_pass` is now the first thing accepted if both `ansible_become_pass` and `ansible_become_password` are used.

s1113950:
@moreati thanks for updating your comments, I've responded to both 👍 

> Additionally I see there are 5 remaining subtasks

I updated the subtasks, now all that's left is to remove the `NOTE` comments and add some documentation (along with a changelog entry)

> `Travis tests are currently red`

There's a commit somewhere in this PR that fixes Travis tests iirc. I'll find it and open a new PR with it.

> I don't know the status of Ansible 2.9 support in current master

There's been multiple github tickets filed of people trying to fix our README with adding `2.9` in the list of supported Ansible versions, so I think we're g2g on it 👍 everything should be good with that version except for collections, but collections weren't really a `first class` feature in Ansible 2.9 (https://github.com/dw/mitogen/pull/715#issuecomment-628753991) so I don't think we need to include collection support in 2.9.

tests/ansible/integration/transport_config/become_pass.yml
moreati: I've adapted #762 to this branch in https://github.com/moreati/mitogen/pull/new/collectionsSupport-tox. The results of the last run were successful. They include the Mitogen Ansible tests, with strategy=linear and strategy=mitogen_linear. It appears same reversal has happened in both, so I'm content this is a correct change.

```
________________________________________________ summary ________________________________________________
  init: commands succeeded
  py27-mode_ansible-ansible2.10-distros_debian-strategy_linear: commands succeeded
  py36-mode_ansible-ansible2.10-distros_debian-strategy_linear: commands succeeded
  py37-mode_ansible-ansible2.10-distros_debian-strategy_linear: commands succeeded
  py27-mode_ansible-ansible2.10-distros_debian: commands succeeded
  py36-mode_ansible-ansible2.10-distros_debian: commands succeeded
  py37-mode_ansible-ansible2.10-distros_debian: commands succeeded
  py27-mode_mitogen-distro_centos7: commands succeeded
  py36-mode_mitogen-distro_centos7: commands succeeded
  py37-mode_mitogen-distro_centos7: commands succeeded
  report: commands succeeded
  congratulations :)
```

mitogen/master.py
moreati: Based on some [adhoc doctests](https://github.com/moreati/mitogen/commit/69ef1980e1dfa223bd63e1e5c2468ce7a996c54c) I'm satisfied with changes to this function.

.ci/ansible_install.py
```diff
@@ -11,7 +11,8 @@
         'pip install '
             '-r tests/requirements.txt '
             '-r tests/ansible/requirements.txt',
-        'pip install -q ansible=={0}'.format(ci_lib.ANSIBLE_VERSION)
+        # encoding is required for installing ansible 2.10 with pip2, otherwise we get a UnicodeDecode error
+        'LC_CTYPE=en_US.UTF-8 LANG=en_US.UTF-8 pip install -q ansible=={0}'.format(ci_lib.ANSIBLE_VERSION)
```
moreati, line 15: I also encountered this, pinning to `pip<20.2` also avoids it https://github.com/dw/mitogen/pull/762/commits/5ecf6488d4dcacf0ab65105b0e19e400f96e7c4f

.ci/azure-pipelines-steps.yml
```diff
@@ -8,23 +8,7 @@ steps:
...
-    # and Azure isn't allowing disabling it apparently: https://developercommunityapi.westus.cloudapp.azure.com/idea/558702/allow-disabling-sip-on-microsoft-hosted-macos-agen.html
-    # the || will activate when running python3 tests
-    # TODO: get python3 tests passing
-    (sudo ln -fs /usr/bin/python$(python.version) /usr/bin/python &&
-      /usr/bin/python -m pip install -U pip wheel setuptools &&
-      /usr/bin/python -m pip install -U virtualenv &&
-      /usr/bin/python -m virtualenv /tmp/venv -p /usr/bin/python$(python.version)) ||
-    (sudo /usr/bin/python$(python.version) -m pip install -U pip wheel setuptools &&
-      /usr/bin/python$(python.version) -m venv /tmp/venv)
     echo "##vso[task.prependpath]/tmp/venv/bin"
```
moreati, line 12: The code that creates this virtualenv has been removed. Is there any point in keeping this line?

.ci/ci_lib.py
```diff
@@ -49,6 +49,10 @@ def have_apt():
     proc = subprocess.Popen('apt --help >/dev/null 2>/dev/null', shell=True)
     return proc.wait() == 0
 
+def have_brew():
+    proc = subprocess.Popen('brew help >/dev/null 2>/dev/null', shell=True)
+    return proc.wait() == 0
+
```
moreati, line 55: I don't think anything in this branch uses this function anymore? Your choice if you want to keep it.

.ci/localhost_ansible_tests.py
moreati: I think this should be one of the test fixes that is split out from this PR.

.ci/localhost_ansible_tests.py
```diff
@@ -20,12 +20,15 @@
...
     os.chmod(KEY_PATH, int('0600', 8))
+    # NOTE: sshpass v1.06 causes errors so pegging to 1.05 -> "msg": "Error when changing password","out": "passwd: DS error: eDSAuthFailed\n", 
+    # there's a checksum error with "brew install http://git.io/sshpass.rb" though, so installing manually
     if not ci_lib.exists_in_path('sshpass'):
-        run("brew install http://git.io/sshpass.rb")
+        os.system("curl -O -L  https://sourceforge.net/projects/sshpass/files/sshpass/1.05/sshpass-1.05.tar.gz && \
+                tar xvf sshpass-1.05.tar.gz && \
+                cd sshpass-1.05 && \
+                ./configure && \
+                sudo make install")
```
moreati, line 31: I think this should be one of the test fixes that is split out from this PR.

.travis.yml
```diff
@@ -18,75 +18,64 @@ cache:
 
 install:
 - grep -Erl git-lfs\|couchdb /etc/apt | sudo xargs rm -v
+- pip install -U pip==20.2.1
```
moreati, line 21: `pip >= 20.2` is what triggers the UnicodeDecodeError you had to work around in ansible_install.py

.travis.yml
```diff
@@ -18,75 +18,64 @@ cache:
...
+    # 2.10; 3.6 -> 2.7
+    # - python: "3.6"
+    #   env: MODE=debops_common VER=2.10.0
+    # 2.10; 2.7 -> 2.7
+    # - python: "2.7"
+    #   env: MODE=debops_common VER=2.10.0
 
     # Sanity check against vanilla Ansible. One job suffices.
-    - python: "2.7"
-      env: MODE=ansible VER=2.8.3 DISTROS=debian STRATEGY=linear
```
moreati, line 51: I would like to keep at least one CI job with `STRATEGY=linear`, as a reference to compare against.

geigi:
Hi :)

I just encountered an issue trying this branch. Here is my setup:
```
ansible 2.10.3
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/venv/ansible/lib/python3.8/site-packages/ansible
  executable location = /home/user/venv/ansible/bin/ansible
  python version = 3.8.6 (default, Sep 25 2020, 09:36:53) [GCC 10.2.0]

```

I did run my playbook like this:

geigi:
Hi :)

I just encountered an issue trying this branch. Here is my setup:
```
ansible 2.10.3
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/venv/ansible/lib/python3.8/site-packages/ansible
  executable location = /home/user/venv/ansible/bin/ansible
  python version = 3.8.6 (default, Sep 25 2020, 09:36:53) [GCC 10.2.0]

```

I did run my playbook like this:
```bash
ANSIBLE_STRATEGY_PLUGINS=/home/user/git/mitogen/ansible_mitogen/plugins/strategy ANSIBLE_STRATEGY=mitogen_linear ansible-playbook -i inventory playbook.yml -l "$HOST$"
```

This is the error message:
```
ERROR! [mux  13615] 09:05:39.967788 E mitogen.[ssh.$DELEGATE_HOST$]: ExternalContext.main() crashed
  File "<stdin>", line 3989, in main
  File "<stdin>", line 3726, in run
  File "<stdin>", line 633, in _profile_hook
  File "<stdin>", line 3715, in _dispatch_calls
  File "<stdin>", line 3661, in _dispatch_one
  File "<stdin>", line 3648, in _parse_request
  File "<stdin>", line 668, in import_module
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1191, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1161, in _load_backward_compatible
  File "<stdin>", line 1508, in load_module
  File "master:/home/user/git/mitogen/ansible_mitogen/target.py", line 85, in <module>
    import ansible_mitogen.runner
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1191, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1161, in _load_backward_compatible
  File "<stdin>", line 1508, in load_module
  File "master:/home/user/git/mitogen/ansible_mitogen/runner.py", line 85, in <module>
    import ansible.module_utils.basic
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1191, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1161, in _load_backward_compatible
  File "<stdin>", line 1508, in load_module
  File "master:/home/user/venv/ansible/lib/python3.8/site-packages/ansible/module_utils/basic.py", line 272, in <module>
    sys.exit(1)
SystemExit: 1
fatal: [$HOST$]: UNREACHABLE! => {"changed": false, "msg": "Channel was disconnected while connection attempt was in progress; this may be caused by an abnormal Ansible exit, or due to an unreliable target.", "unreachable": true}
```

The task where the error occurred is a `delegate_to` task to host `$DELEGATE_HOST$` (Debian 8, Python 2.7.9 and Python 3.4.2). Just before the task where the crash occured, exactly the same task is executed fine to a different host (Debian 7, Python 2.7.3, no python3). 

Thanks for your work so far and let me know if you need more information!

.ci/localhost_ansible_tests.py
s1113950: The old sshpass version doesn't work anymore on the test setup, so I'm fine splitting it out but then that PR would need to land before this one does. WDYT?

.travis.yml
s1113950: Ok, nice find! I'm in favor of passing the encoding on command install rather than pinning to an older version of pip, so the fix isn't lost in the future. I'm imagining a scenario where a new pip version comes along that we want, and then we try and upgrade to it and run into the potential UnicodeDecodeError again, but forget how to resolve it. For that reason, I'm inclined to keep the encoding as-is in ansible-install.py. WDYT?

.travis.yml
s1113950: I migrated the linear strategy test to the azure tests section because travis doesn't let us have long-running jobs. It caps them at 50 min and the new Ansible version tests take like 1.5 hours on vanilla Ansible :( The comment I left points to a github comment explaining the 50 min limit, but I'll make the comment more clear 👍

s1113950:
@geigi thanks for testing! :)

Your issue happens because Ansible itself threw an system exit, the relevant code is here:
https://github.com/ansible/ansible/blob/v2.10.3/lib/ansible/module_utils/basic.py#L264-L272

Setting `ansible_python_interpreter: auto_legacy` told Mitogen/Ansible how to detect which python version to run with. What I believe happened was that without setting `auto_legacy`, a bad Python version was detected and Ansible 💥 which caused Mitogen to 💥 .
Mitogen defaults to the `auto` method if it can't determine what Python version to use, instead of `auto_legacy`.

Also, I made a different ticket [here](https://github.com/dw/mitogen/issues/765) because I think if possible, Mitogen should error nicer and say Ansible 💥 instead of throwing a massive traceback.

geigi:
Thanks for the explanation 😊 
One thing is not clear to me yet:

Running the same playbook without mitogen and without the `auto_legacy` option worked just fine and no system exit is thrown by ansible. Why does the system exit only occur when using mitogen?

s1113950:
From what I can tell, looks like `Ansible` tries `auto_legacy` mode first when detecting an interpreter: https://github.com/ansible/ansible/blob/v2.10.0/lib/ansible/executor/interpreter_discovery.py#L115  which would explain why it doesn't system exit but `Mitogen` does (`Mitogen` defaults to `auto` instead of `auto_legacy`).

Since `Mitogen` relies on [sending the interpreter discovery mode to `Ansible`](https://github.com/dw/mitogen/blob/master/ansible_mitogen/transport_config.py#L119) and `auto_legacy` is the legacy mode of detecting an interpreter, `Mitogen` defaults to the `auto` way of detecting an interpreter. `Ansible` throws a `sys.exit(1)` if the detected python version is too old/unsupported, and that big traceback you found is what `Mitogen` currently does if `Ansible` throws a `sys.exit(1)`.

geigi:
Good to know that the default for `ansible_python_interpreter` is different with mutagen :) Thanks!

s1113950:
@moreati I made changes/commented on your latest review. I'm going to make an `0.3.0-rc.0` branch now with this PR in it, as well as a `0.2-release` branch and tag it with `0.2.10-rc.0` that will contain everything in `master` except for this PR in it as discussed previously in the PR thread. Can't make a `0.2.10` tag yet because tests need fixing on `master` with some of what's in this PR. There's still more to do here as well (docs, changelog entries I forgot to make, somehow testing bug [here](https://github.com/dw/mitogen/issues/766), etc), but I'd love to get something merged before Dec 31st 2020 because ⬇️ ⬇️ ⬇️ 

PSA: travis.org is being EOL EOY, which will affect Mitogen since our `required` tests are on travis.org. I don't have access to the `settings` of Mitogen, so I can't change this. @dw can you point tests to travis.com instead? See [migration from travis.org to travis.com](https://docs.travis-ci.com/user/migrate/open-source-repository-migration#q-when-will-the-migration-from-travis-ciorg-to-travis-cicom-be-completed)

s1113950:
New releases made: https://github.com/dw/mitogen/releases 

TL;DR of this PR thread: Mitogen has 2 releases now due to a bunch of API-breaking changes Ansible 2.10 made, 1 with Ansible < 2.10 and one Ansible 2.10+. Please tag any Github issues related to either release candidate with the new labels `affects-0.2` or `affects-0.3` 🙏

jfcabral:
Thank you for you hard work! It’s great to see mitogen on the verge of supporting Ansible v2.10! 👍

Issue #720, opened 2020-06-04
RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode


lonix1:
## environment:
Ubuntu 20.04
Python 3.8.2
Ansible 2.9.6
Mitogen 0.2.9 (also tried latest prerelease zip from github)

## playbook:
````
- name: foo
  hosts: localhost
  connection: local
  gather_facts: 'true'
````

## result:
> TASK [Gathering Facts] ***********************************************************************************************************************************************************************
/usr/lib/python3.8/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  return io.open(fd, *args, **kwargs)
ok: [localhost]

This warning did not show before I upgraded to Ubuntu 20.04. I assume the change is due to a newer python version?

Frankly it's just a warning so I don't mind it. But I thought the mitogen devs may want to know about it.

**Comment by s1113950** [comment]
Thanks for the ticket!

**Comment by BirkhoffLee** [comment]
This is still an issue:

```
- name: File Synchronisation
  ansible.posix.synchronize:
    src: .
    dest: /path/to/dest
```

```
TASK [File Synchronisation] *************************************************************************************************
/usr/local/Cellar/python@3.9/3.9.9/Frameworks/Python.framework/Versions/3.9/lib/python3.9/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  return io.open(fd, *args, **kwargs)
changed: [SFM-VPS-2]
```

Issue #691, opened 2020-02-19
Runtime warnings on Python 3.8+ local connection


tprasadtp:
Is there any reason for using buffering in `fdopen()` or `open()`? IT generates a runtime warning when connection is local.

Steps to reproduce
```python
import mitogen.master                          

In [2]: r = mitogen.master.Router()                                             

In [3]: conn = r.local()                                                        
/usr/lib/python3.8/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  return io.open(fd, *args, **kwargs)
```

igormcsouza:
Having the same Warning which bothers me to hell!!

PR #739, opened 2020-08-26
Fixes RuntimeWarning for improper buffering on local connection in Python 3.8+

MetricMike:
```bash
[mux  24512] 18:54:08.981562 D mitogen.io: EpollPoller.poll(29.704136800002743)
/home/metricmike/.asdf/installs/python/3.8.5/lib/python3.8/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  return io.open(fd, *args, **kwargs)
```
Change in Python 3.8 is described at https://bugs.python.org/issue32236
tl;dr - python used to silently ignore `open(fd, 'rb', 1)` on binary mode, now they don't because it could potentially cause issues

During the Python3 port, mitogen switched a file descriptor at (https://github.com/dw/mitogen/commit/410016ff479b717b2066cb3e1c673658d1ab43fe#diff-3540c3617f4566023f000914cebe2d71L1764-R1831) from text to binary mode on this commit, but left line buffering on. As far as I can tell, this is the only file descriptor where both the mode and buffering option didn't remain consistent on that commit.
image

When building from this commit, the warning goes away and my not super trivial playbook runs successfully so this should solve issues #691 and #720, but I ran into several issues while building/testing and would appreciate some help verifying:
- Python 3.8 breaks at least the CFFI version used in `tests/requirements.txt`. Bumping that to 1.14.2 allows `.ci/localhost_ansible_install.py` to finish.
- Ubuntu 18.04 and up remove /usr/bin/python, so I needed to alias that to python3 in order to get `.ci/localhost_ansible_tests.py` to partially pass
- brew is expected but not installed for `.ci/localhost_ansible_tests.py`
- `.ci/*` expect a clean machine and I was building/testing locally

### DIFF
--- a/mitogen/core.py
+++ b/mitogen/core.py
@@ -3860,7 +3860,7 @@ class ExternalContext(object):
         else:
             core_src_fd = self.config.get('core_src_fd', 101)
             if core_src_fd:
-                fp = os.fdopen(core_src_fd, 'rb', 1)
+                fp = os.fdopen(core_src_fd, 'rb', 0)
                 try:
                     core_src = fp.read()
                     # Strip "ExternalContext.main()" call from last line.


s1113950:
`.ci/localhost_ansible_tests.py` is only used for Mac tests, not too sure why but it was that way before I started on the project and I haven't had the time to standardize the test framework since. 

Unfortunately, Mitogen tests don't really work super well ran locally. Definitely something that needs fixing up! What I've done in the meantime is make my own little test lib here: https://github.com/s1113950/mitogen-test where I run a certain `.yml` file in a container that gets created on the fly (or locally if desired). It's not perfect but it's faster for rapid dev than waiting for Mitogen's whole test suite to finish.

s1113950:
I got the same error you got on another PR! https://github.com/dw/mitogen/pull/715 --> https://dev.azure.com/dw-mitogen/Mitogen/_build/results?buildId=985&view=logs&j=91b5f601-2ef2-5ad3-ce48-a74a8d70b8fa&t=102d5197-072a-520e-94dd-1966b8f15f2b 

This one specifically:
```
File "/usr/lib/python2.7/posixpath.py", line 73, in join
    path += '/' + b
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe3 in position 25: ordinal not in range(128)
```
when I figure it out I'll let you know

s1113950:
@MetricMike try setting these env vars: `LC_CTYPE=en_US.UTF-8 LANG=en_US.UTF-8`

moreati:
I just encountered the same error while expanding coverage of Tox tests (follow on to #762). Changing the same line does remove the repeated `RuntimeWarning`s. I don't pretend to understand the file descriptor juggling that Mitogen does, and I don't know if buffering=0, as opposed to default buffering is the right thing.

moreati:
I'd like to merge #775 before this

moreati:
@MetricMike please could you update this to current master. Otherwise, may I update this branch on your behalf?

MetricMike:
rebase failed - will try to address this weekend but otherwise feel free to update :)

MetricMike:
I missed the 0.2.10 / 0.3.0 split. Latest commits should pass CI.

Can this get added to #800 as well?

moreati:
Shouldn't be a problem, sorry for the wait.

PR #743, opened 2020-09-18
WIP: Fixes: Mitogen fails if file is changed during transfer and validate_checksum is false

soisetu:
Fixes: Mitogen fails if file is changed during transfer and validate_checksum is false

  docs/api.rst.

### DIFF
--- a/ansible_mitogen/connection.py
+++ b/ansible_mitogen/connection.py
@@ -1003,7 +1003,7 @@ class Connection(ansible.plugins.connection.ConnectionBase):
         )
         return rc, stdout, stderr
 
-    def fetch_file(self, in_path, out_path):
+    def fetch_file(self, in_path, out_path, validate_checksum=True):
         """
         Implement fetch_file() by calling the corresponding
         ansible_mitogen.target function in the target.
@@ -1018,7 +1018,8 @@ class Connection(ansible.plugins.connection.ConnectionBase):
             context=self.context,
             # in_path may be AnsibleUnicode
             in_path=mitogen.utils.cast(in_path),
-            out_path=out_path
+            out_path=out_path,
+            validate_checksum=validate_checksum,
         )
 
     def put_data(self, out_path, data, mode=None, utimes=None):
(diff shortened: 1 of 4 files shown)


soisetu:
Hi @s1113950!

We just changed to Mitogen. Our log fetching playbook started to fail pretty often after the change. I tracked it to two things: we weren't using `validate_checksum: false` with fetch and that Mitogen did not care about it.

So what do you think about fix? I will make tests etc. if this is okay for you.

s1113950:
Hello! 👋 I like it 😄 Thanks for the fix! I took a look at your test error and I think it's unrelated to the change you made (and have fixed the encoding error in a different PR)

soisetu:
I tried to do test for this but did not figure how to inject data to received file. If you can help me to make test I can do it. Otherwise you can merge this.

s1113950:
Can you make a test in the `regression` test dir that uses `validate_checksum: false` https://docs.ansible.com/ansible/latest/modules/fetch_module.html#parameter-validate_checksum ? I'm curious now, how did you run into this issue? Do you inject data into the received file on your end? 😄

soisetu:
Our small regression test environment is about 10 VMs. After tests are run, logs are collected. One of the collected log files is `/var/log/messages`. Since it is live system, it is very possible that at least in one VM something writes to file in the middle of transfer. This was issue also without mitogen, but appeared less often: Ansible copies file first and transfer it after and mitogen streams the original file.

I hadn't noticed this issue before mitogen. It was coincidence that it happened when I changed log collector playbook back to linear strategy. After that I added `validate_checksum: false` to every fetch `linear` strategy was working but `mitogen_linear` was not. So I started digging mitogen's code.

I tried to do test in file_service_tests via injecting method to Receiver class. But since none of my code was run I think tests actually run another process? What I was trying to do was to duplicate messages so original checksum and received checksum should be different and with `validate_checksum: false` that should be okay. So I think I cannot make test in regression tests folder since there should be some process writing to file being fetched.

s1113950:
Instead of `file_service_test.py` try adding another test func to this class:
https://github.com/dw/mitogen/blob/master/tests/ansible/tests/connection_test.py#L112 :)
there you can pass in the new `validate_checksum` value for `fetch_file`

PR #757, opened 2020-11-08
ssh: Match newer ssh host key prompt that accepts the fingerprint

### LINKED ISSUE #75: importer: provide hints about native module dependencies
Some Python modules comes with speedup modules that we don't support transferring. Not sure if there will ever be a strong enough use case for supporting remote import of Python extensions, so instead, ModuleResponder should provide negative hints to Importer about dependencies in a Python module that are satisfied by an extension module in the master process.

Avoids roundtrips for these situations:

12:39:17.611535 W mitogen: ModuleFinder(): cannot find source for 'markupsafe._speedups'
12:39:17.822373 E mitogen: _build_tuple('_imaging_gif'): could not locate source

And plenty more just those

moreati:
This fixes an ERROR in test_accept_enforce_host_keys() while running the
test suite.

Fixes #75

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -21,7 +21,8 @@ v0.2.10 (unreleased)
 To avail of fixes in an unreleased version, please download a ZIP file
 `directly from GitHub <https://github.com/dw/mitogen/>`_.
 
-*(no changes)*
+* :gh:issue:`756` ssh connections with `check_host_keys='accept'` would
+  timeout, when using recent OpenSSH client versions.
 
 
 v0.2.9 (2019-11-02)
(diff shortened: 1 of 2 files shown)


s1113950 (approved):

s1113950:
🎉 nice fix. Unrelated, but since you're another main contributor to Mitogen could you give https://github.com/dw/mitogen/pull/715 some 👀 :D 🙏  ? It should be ready at this point, at least as an initial release (there are bound to be some edge cases but it passes tests and is a good stepping stone imo). I fixed tests in that PR; will merge your change now.

moreati:
Thank you, I'll take a look at #715, but be aware Mitogen is still mostly a black box of threading, signals, and other voodoo to me.

PR #769, opened 2020-12-29
Add ansible.legacy.setup to be fixed on py3.5

MarkusTeufelberger:
This fix was already mentioned in https://github.com/dw/mitogen/pull/715#issuecomment-731601226 but somehow seems to have not made it into the final version.

This showed up as an actual issue on an Ubuntu 16.04 host for me using `ansible-base==2.10.4` and `mitogen==0.3.0-rc.0`.


s1113950:
Thanks for catching this!

PR #774, opened 2021-01-10
tests: Correct DISTRO vs DISTROS usage in CI configurations

moreati:
The environment variable `DISTRO` is used to set the target Docker image
used when running the Mitogen unit tests. `DISTROS` is used for the
Ansible integration tests. `VER` sets the version of Ansible that is
installed on the controller.

When `MODE=Mitogen`
- there is no need to set `VER`, because nothing that installs Ansible
- it does not make sense to set `DISTROS`.

### DIFF
--- a/.ci/ci_lib.py
+++ b/.ci/ci_lib.py
@@ -149,9 +149,12 @@ class Fold(object):
 
 os.environ.setdefault('ANSIBLE_STRATEGY',
     os.environ.get('STRATEGY', 'mitogen_linear'))
+# Ignoreed when MODE=mitogen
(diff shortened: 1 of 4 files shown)


s1113950 (approved):

.ci/ci_lib.py
```diff
@@ -217,9 +217,12 @@ def __exit__(self, _1, _2, _3):
 
 os.environ.setdefault('ANSIBLE_STRATEGY',
     os.environ.get('STRATEGY', 'mitogen_linear'))
+# Ignoreed when MODE=mitogen
```
s1113950, line 220: typo, but not enough to warrant delaying this imo

PR #775, opened 2021-01-10
tests: Add Python 3.9 coverage

moreati:


moreati:
Travis Python/Linux availability are documented at https://docs.travis-ci.com/user/languages/python/#python-versions

moreati:
Interesting looking quotes from Python What's New documents

From https://docs.python.org/3/whatsnew/3.8.html

>The subprocess module can now use the os.posix_spawn() function in some cases for better performance. Currently, it is only used on macOS and Linux (using glibc 2.24 or newer) if all these conditions are met:
>-    close_fds is false;
>-   preexec_fn, pass_fds, cwd and start_new_session parameters are not set;
>-    the executable path contains a directory.

> subprocess.Popen can now use os.posix_spawn() in some cases for better performance. On Windows Subsystem for Linux and QEMU User Emulation, the Popen constructor using os.posix_spawn() no longer raises an exception on errors like “missing program”. Instead the child process fails with a non-zero returncode. (Contributed by Joannah Nanjekye and Victor Stinner in bpo-35537.)

From https://docs.python.org/3/whatsnew/3.9.html

> Previously, sys.stderr was block-buffered when non-interactive. Now stderr defaults to always being line-buffered.

> The sys.getcheckinterval() and sys.setcheckinterval() functions have been removed. They were deprecated since Python 3.2. Use sys.getswitchinterval() and sys.setswitchinterval() instead.

> __import__() and importlib.util.resolve_name() now raise ImportError where it previously raised ValueError. Callers catching the specific exception type and supporting both Python 3.9 and earlier versions will need to catch both using except (ImportError, ValueError):

> The select.epoll.unregister() method no longer ignores the EBADF error. (Contributed by Victor Stinner in bpo-39239.)

moreati:
Test failure I'm hunting, occurs on Python 3.8 and Python 3.9

```
FAIL: test_stats (responder_test.ForwardTest)
  File "/home/alex/src/mitogen/tests/responder_test.py", line 193, in test_stats
    self.assertEquals(2+os_fork, self.router.responder.get_module_count)
AssertionError: 2 != 3

```

moreati:
Huh, wouldn't have guessed that

Python 3.7
```
test_stats (responder_test.ForwardTest) ... ['mitogen.parent', 'plain_old_module']
ok
```

Python 3.8
```
test_stats (responder_test.ForwardTest) ... ['mitogen.parent', 'msvcrt', 'plain_old_module']
FAIL
```
<details>
<summary>Diagnostics diff</summary>

```diff
diff --git a/mitogen/master.py b/mitogen/master.py
index e54795cb..663cb33b 100644
--- a/mitogen/master.py
+++ b/mitogen/master.py
@@ -919,6 +919,7 @@ class ModuleResponder(object):
         self.good_load_module_size = 0
         #: Number of negative LOAD_MODULE messages sent.
         self.bad_load_module_count = 0
+        self.got = []
 
         router.add_handler(
             fn=self._on_get_module,
@@ -1085,6 +1086,7 @@ class ModuleResponder(object):
 
         fullname = msg.data.decode()
         self._log.debug('%s requested module %s', stream.name, fullname)
+        self.got.append(fullname)
         self.get_module_count += 1
         if fullname in stream.protocol.sent_modules:
             LOG.warning('_on_get_module(): dup request for %r from %r',
diff --git a/tests/responder_test.py b/tests/responder_test.py
index 2653589c..639b3294 100644
--- a/tests/responder_test.py
+++ b/tests/responder_test.py
@@ -190,6 +190,7 @@ class ForwardTest(testlib.RouterMixin, testlib.TestCase):
 
         os_fork = int(sys.version_info < (2, 6))
         self.assertEquals(256, c2.call(plain_old_module.pow, 2, 8))
+        print(self.router.responder.got)
         self.assertEquals(2+os_fork, self.router.responder.get_module_count)
         self.assertEquals(2+os_fork, self.router.responder.good_load_module_count)
         self.assertLess(10000, self.router.responder.good_load_module_size)
```
</details>

moreati:
Gotcha. Commit https://github.com/python/cpython/commit/880d42a3b247306f67837aa95e23f7c3471a30a3 (first released in Python 3.8a0) moved an import of `msvcrt` from an `if <win32>` block, into a `try/except` block. So now the import is tried even on Linux or MacOS.

mitogen/core.py
```diff
@@ -1269,6 +1269,13 @@ class Importer(object):
         # a negative round-trip.
         'builtins',
         '__builtin__',
+
+        # On some Python releases (e.g. 3.8, 3.9) the subprocess module tries
+        # to import of this Windows-only builtin module.
+        'msvcrt',
+
+        # Python 2.x module that was renamed to _thread in 3.x.
+        # This entry avoids a roundtrip on 2.x -> 3.x.
```
moreati, line 1278: David, this entry dates back to https://github.com/dw/mitogen/commit/d6dcb8d010de815840b1a3e53a69fd2f9b4a34de. Do you remember if it was for correctness, or purely an optimisation?

moreati:
Catch 22. PRs require mitogen-hq.mitogen (Linux Mito37Debian_27) and  mitogen-hq.mitogen (Linux Ansible_210_35) to pass, before they're merged. This PR changes the name of those jobs.

moreati:
> Catch 22. PRs require mitogen-hq.mitogen (Linux Mito37Debian_27) and mitogen-hq.mitogen (Linux Ansible_210_35) to pass, before they're merged. This PR changes the name of those jobs.

Perhaps the answer is to make "mitogen-hq.mitogen" required, but not the individual Azure jobs (e.g. mitogen-hq.mitogen (Linux Ansible_210_27). If azur could be convinced to make the top level one depend on all its child jobs.

Issue #777, opened 2021-01-11
0.3.0-rc.0 release tarball from github reports itself as 0.2.9.


wilhelmy:
I just installed 0.3.0-rc.0 and it reports itself as 0.2.9.

```
fukushima ~/s/t/mitogen-0.3.0-rc.0 $ pip install . 
Processing /home/mw/src/tar/mitogen-0.3.0-rc.0
Using legacy 'setup.py install' for mitogen, since package 'wheel' is not installed.
Installing collected packages: mitogen
    Running setup.py install for mitogen ... done
Successfully installed mitogen-0.2.9
fukushima ~/s/t/mitogen-0.3.0-rc.0 $ python ./setup.py --version
0.2.9
```

s1113950:
🤦 good catch, thanks for filing a ticket

PR #783, opened 2021-01-18
fix 0.3.0-rc.0 version reporting

s1113950:
Fixes https://github.com/mitogen-hq/mitogen/issues/777

### DIFF
--- a/mitogen/__init__.py
+++ b/mitogen/__init__.py
@@ -35,7 +35,7 @@ be expected. On the slave, it is built dynamically during startup.
 
 
 #: Library version as a tuple.
-__version__ = (0, 2, 9)
+__version__ = (0, 3, "0-rc.0")
 
 
 #: This is :data:`False` in slave contexts. Previously it was used to prevent
--- a/mitogen/__init__.py
+++ b/mitogen/__init__.py
@@ -35,7 +35,7 @@ be expected. On the slave, it is built dynamically during startup.
 
 
 #: Library version as a tuple.
-__version__ = (0, 3, "0-rc.0")
+__version__ = (0 , 3, 0, 'rc', 0)
 
 
 #: This is :data:`False` in slave contexts. Previously it was used to prevent


moreati (approved):

moreati:
In setup.py `__version__` is has `eval()` called on it. I don't think it's a problem, but perhaps note worthy.

https://github.com/mitogen-hq/mitogen/blob/6e0a01f8a04b227c918b31fe3c29720357661bc2/setup.py#L38-L40

moreati:
How do a sample of projects do their version, wrt to alphas/betas and int vs string?

- Mitogen: all previous releases have used a 3-tuple of integers. There haven't been any alphas or betas.
  Based `git log -G __version__ -p mitogen/__init__.py`, and https://pypi.org/project/mitogen/#history

- Python
   ```
  >>> sys.version_info
  sys.version_info(major=3, minor=10, micro=0, releaselevel='alpha', serial=4)
  >>> sys.version
  '3.10.0a4 (default, Jan  5 2021, 00:08:30) \n[GCC 9.3.0]'
  >>> print(sys.version)
  3.10.0a4 (default, Jan  5 2021, 00:08:30) 
  [GCC 9.3.0]
  ```

- Ansible (now ansible-base, as a string in ansible/release/__init.py)
  ```
  >>> ansible.__version__
  '2.10.0b1'
  ```
- Hypothesis: has always used 3-tuple of integers.
  Based on `git log -P -G ^__version_info__ -p`, and https://pypi.org/project/hypothesis/#history

mitogen/__init__.py
```diff
@@ -35,7 +35,7 @@
 
 
 #: Library version as a tuple.
-__version__ = (0, 2, 9)
+__version__ = (0, 3, "0-rc.0")
```
moreati, line 38: Suggestion: follow the convention of Python's `sys.version_info` and Django, keeping the first 3 elements as integers. E.g.

```python
__version__ = (0 ,3, 0, 'rc', 0)
```

s1113950:
> In setup.py `__version__` is has `eval()` called on it. I don't think it's a problem, but perhaps note worthy.
> 
> https://github.com/mitogen-hq/mitogen/blob/6e0a01f8a04b227c918b31fe3c29720357661bc2/setup.py#L38-L40

I'm in favor of removing the way Mitogen does versioning in place of [versioneer](https://github.com/python-versioneer/python-versioneer). I've used it before in other projects and it automatically detects version based on things like git tag. I see that it says it supports python3 only at the moment; if we still want to support python2 then we can use https://github.com/python-versioneer/python-versioneer/tree/0.18 which supports back to python2.6. If we migrated to versioneer then there'd be no need for either the custom `version` function in `setup.py` or having to maintain a tuple ourselves :D

moreati:
Notes to self
- The git tag `v0.3.0-rc.0` already exists
- `__version__ = (0, 3, "0-rc.0")` results in package `mitogen-0.3.0rc0` (no dash before "rc0")
- [Versioneer](https://github.com/python-versioneer/python-versioneer)
  - uses git tags, or version embedded in parent directory name (typical for unpacked sdist), e.g.
    - `mitogen-0.3.0rc0+20.gdb354df3.dirty`
    - `mitogen-0.3.0rc0+21.gac720689`
    - `mitogen-0.3.0rc1 (after `git tag -a v0.3.0rc1`, note the `v`, tag now deleted)
  - has no dependencies of it's own
  - doesn't introduce runtime, or CI dependencies (90% confidence)
  - generates
    - `versioneer.py` (imported by `setup.py`, checked in file and sdist bundled file are identical)
    - `mitogen/_verison.py` (imported in `mitogen/__init__.py`, checked in file introspects environment, sdist bundled file de-serializes embedded json)
  - works with string versions, not tuple style versions
  - relies on executing code at import time to populate `mitogen.__version__`
  - would need some remedial work in the minifier, and test suite
     ```
    ERROR: test_minify_all (minify_test.MitogenCoreTest)
      File "/home/alex/src/mitogen/tests/minify_test.py", line 112, in test_minify_all
        self._test_syntax_valid(minified, name)
      File "/home/alex/src/mitogen/tests/minify_test.py", line 71, in _test_syntax_valid
        compile(minified, name, 'exec')
      File "mitogen/_version.py", line 37
        def get_config():
          ^
    IndentationError: expected an indented block

    FAIL: test_simple (module_finder_test.FindRelatedTest)
      File "/home/alex/src/mitogen/tests/module_finder_test.py", line 295, in test_simple
        self.assertEquals(set(related), self.SIMPLE_EXPECT)
    AssertionError: Items in the first set but not the second:
    u'mitogen._version'

    FAIL: test_stats (responder_test.ForwardTest)
      File "/home/alex/src/mitogen/tests/responder_test.py", line 194, in test_stats
        self.assertEquals(2+os_fork, self.router.responder.good_load_module_count)
    AssertionError: 2 != 3
    ```
  - has the meme-tastic [`long_description=LONG`](https://github.com/python-versioneer/python-versioneer/blob/4386af615ef66ff10b8759b0b696b696828ae68e/setup.py#L159) in its setup.py :smile:

dw:
This sounds fine as a built-time requirement (loving it's a single module) assuming that the `tar.gz` releases it produces do not require Git to be installed on the target machine (I think this is the case?)

On the compatibility front, while the core library treats Python 2 essentially as a stable near-universal network protocol for manipulating remote infrastructure, the terms /could/ be a little looser for the host machine, where it's much simpler to ensure a modern Python build is available (even in crusty weird old environments). I'm wary of taking little sips of the Python 3 kool-aid though, as it could lead to painting the project into an unforeseen corner. The idea of Python 2 disappearing any time in the coming years is some adorable meme that in no way reflects the real world lifecycle of enterprise infrastructure.

moreati:
> assuming that the tar.gz releases it produces do not require Git to be installed

Correct

> I'm wary of taking little sips of the Python 3 kool-aid though, as it could lead to painting the project into an unforeseen corner

Versioneer 0.18 would require Python 2.6+ on the Mitogen initiator (maybe also child nodes), I think we can do without the 0.19 features.

moreati:
I'm preparing a 0.3.0rc1 and 0.2.10rc1 as PRs. Propose to make a decision on versioneer in 0.3.0rc2.

s1113950:
> I'm preparing a 0.3.0rc1 and 0.2.10rc1 as PRs. Propose to make a decision on versioneer in 0.3.0rc2.

Sounds good 👍 

> This sounds fine as a built-time requirement (loving it's a single module) assuming that the tar.gz releases it produces do not require Git to be installed on the target machine (I think this is the case?)

Correct, iirc nothing needs to be installed on the target machine (at least I haven't needed to on the versions I've used for my own projects).

mitogen/__init__.py
s1113950: Updated! I just tested and a `pip install .` on the repo after changing `__version__` to be what you suggested resulted in the same version string as before.

mitogen/__init__.py
```diff
@@ -35,7 +35,7 @@
 
 
 #: Library version as a tuple.
-__version__ = (0, 2, 9)
+__version__ = (0 , 3, 0, 'rc', 0)
```
moreati, line 38: This results in the version string "0.3.0.rc.0", which setuptools/pip normalises to "0.3.0rc0", to meet PEP440. For #801 I tweaked setup.py as well https://github.com/mitogen-hq/mitogen/pull/801/commits/9b1617f26f03d6b90065bbdb71412c216f3c2e1a

As far as the outside world is concerned both are good, so LGTM.

Issue #770, opened 2021-01-04
AttributeError: 'PluginLoader' object has no attribute 'get_with_context'


arrfab:
When using mitogen from master branch, ansible now tracebacks with the following error :

```
ansible -m ping all -vvv
ansible 2.9.16
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.6.8 (default, Aug 24 2020, 17:57:11) [GCC 8.3.1 20191121 (Red Hat 8.3.1-5)]

ERROR! Unexpected Exception, this is probably a bug: 'PluginLoader' object has no attribute 'get_with_context'
the full traceback was:

Traceback (most recent call last):
  File "/usr/bin/ansible", line 123, in <module>
    exit_code = cli.run()
  File "/usr/lib/python3.6/site-packages/ansible/cli/adhoc.py", line 166, in run
    result = self._tqm.run(play)
  File "/usr/lib/python3.6/site-packages/ansible/executor/task_queue_manager.py", line 261, in run
    strategy = strategy_loader.get(new_play.strategy, self)
  File "/usr/lib/python3.6/site-packages/ansible/plugins/loader.py", line 552, in get
    self._module_cache[path] = self._load_module_source(name, path)
  File "/usr/lib/python3.6/site-packages/ansible/plugins/loader.py", line 525, in _load_module_source
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/var/lib/tobisna-dev/mitogen-master/ansible_mitogen/plugins/strategy/mitogen_linear.py", line 55, in <module>
    import ansible_mitogen.loaders
  File "/var/lib/tobisna-dev/mitogen-master/ansible_mitogen/loaders.py", line 62, in <module>
    connection_loader__get = connection_loader.get_with_context
AttributeError: 'PluginLoader' object has no attribute 'get_with_context'

```
It happens on CentOS 8, with ansible-2.9.16-1.el8.noarch , using default python 3.6.8 and mitogen from current master branch.
But it was working some days ago, so I just tried to revert to specific tag https://github.com/dw/mitogen/releases/tag/v0.2.10-rc.0 and then it worked again.

s1113950:
[...]

PR #785, opened 2021-01-18
Checks Ansible version before loaders are loaded

s1113950:
Fixes https://github.com/mitogen-hq/mitogen/issues/770

This moves checking the supported Ansible version to `loaders.py` because `loaders` are imported before `strategy`. Bug happened  when an older Ansible version was used and `connection_loader__get = connection_loader.get_with_context` failed, as that used to be called `connection_loader.get`.

This PR also removes some cruft because we're no longer supporting super old Ansible versions.

### DIFF
--- a/ansible_mitogen/loaders.py
+++ b/ansible_mitogen/loaders.py
@@ -31,6 +31,7 @@ Stable names for PluginLoader instances across Ansible versions.
 """
 
 from __future__ import absolute_import
+import distutils.version
 
 __all__ = [
     'action_loader',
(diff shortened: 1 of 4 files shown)


extmind:
Maybe update the URLs in `NEW_VERSION_MSG` while you're moving it.

ansible_mitogen/loaders.py
```diff
@@ -41,21 +42,59 @@
...
+    if v[:2] > ANSIBLE_VERSION_MAX:
+        raise ansible.errors.AnsibleError(
+            NEW_VERSION_MSG % (ansible.__version__, ANSIBLE_VERSION_MAX)
+        )
+
+
+# this is the first file our strategy plugins import, so we need to check this here
+# in prior Ansible versions, connection_loader.get_with_context didn't exist, so if a user
+# is trying to load an old Ansible version, we'll fail and error gracefully
+assert_supported_release()
```
moreati, line 89: I'm worried this changes any API guarantee (assuming mitogen_ansible has one) too much. E.g. moving the AnsibleErrors that a consumer would have to handle from runtime to import time

s1113950: I thought about that too. `strategy.py` imports `loaders.py` tho, so an end user would run into the AnsibleError in either case regardless I think? 🤔 unless they are just loading `loaders.py` but that doesn't really do anything other than monkeypatch imports.

moreati (changes requested):
I agree on Ansible 2.10+ only for 0.3.x. This just needs a changelog entry, then I'm happy to merge.

moreati (approved):

s1113950:
Test failed because `Error response from daemon: toomanyrequests: Rate exceeded` CI rate limiting 😞

moreati:
> Test failed because `Error response from daemon: toomanyrequests: Rate exceeded` CI rate limiting disappointed

I saw. I think my implementation of the sleeps is wrong, each pull is running in parallel so the sleep is a noop.

On the Checks tab of each PR expand Azure Pipelines. Each failed job will have a Re-run link. I'm clicking it now
image

PR #787, opened 2021-01-20
Correct version checks of Ansible and Python

### LINKED ISSUE #786: mitogen_ansible compares ansible.__version__ without parsing it
`ansible.__version__` is a string that contains the release version of Ansible, as a string (e.g. `"2.10.0"`), ansible_mitogen contains several checks against this to decide on behaviours. These worked up to and including Ansible 2.9, but they are incorrect for 2.10 onward.

I have a patch in the works.

moreati:
Checks were performed using character by character string comparisons. This gives incorrect results when checking Ansible 2.10, because '2.10' < '2.9'.

Fixes #786

### DIFF
--- a/ansible_mitogen/mixins.py
+++ b/ansible_mitogen/mixins.py
@@ -226,7 +226,7 @@ class ActionModuleMixin(ansible.plugins.action.ActionBase):
         with a pipelined call to :func:`ansible_mitogen.target.prune_tree`.
         """
         LOG.debug('_remove_tmp_path(%r)', tmp_path)
-        if tmp_path is None and ansible.__version__ > '2.6':
+        if tmp_path is None and ansible_mitogen.util.ansible_version[:2] > (2, 6):
             tmp_path = self._connection._shell.tmpdir  # 06f73ad578d
         if tmp_path is not None:
             self._connection.get_chain().call_no_reply(
@@ -335,7 +335,7 @@ class ActionModuleMixin(ansible.plugins.action.ActionBase):
     def _set_temp_file_args(self, module_args, wrap_async):
         # Ansible>2.5 module_utils reuses the action's temporary directory if
         # one exists. Older versions error if this key is present.
-        if ansible.__version__ > '2.5':
+        if ansible_mitogen.util.ansible_version[:2] > (2, 5):
             if wrap_async:
                 # Sharing is not possible with async tasks, as in that case,
                 # the directory must outlive the action plug-in.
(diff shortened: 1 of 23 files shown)


s1113950 (changes requested):
Future Mitogen will only support Ansible 2.10+, so we can remove old checks against old versions of Ansible :)

moreati:
I don't want to remove Ansible 2.8 & 2.9 support from master permanently. I'm working on a branch to restore it.

s1113950:
🤔 Ansible 2.8 was EOL a week ago: https://access.redhat.com/support/policy/updates/ansible-engine ; also 2.8's API is quite different from 2.10+. What's the motivation for supporting a dead Ansible version?

moreati:
> thinking Ansible 2.8 was EOL a week ago: https://access.redhat.com/support/policy/updates/ansible-engine ; also 2.8's API is quite different from 2.10+.

and even Ansible 2.10 had it's final release in the last few days https://groups.google.com/g/ansible-announce/c/6_AmfDp2fCw. Ansible 2.9 might hang around longer, for RHEL customers?

> What's the motivation for supporting a dead Ansible version?

How I imagined migrating to Mitogen 0.3 for my projects, in 2 change-one-thing-at-a-time steps, e.g.
1. Upgrade to Mitogen 0.3.x, keeping Ansible 2.y
2. Upgrade to Ansible 2.y+z, keeping Mitogen 0.3.x

I'm less enamoured with the idea of supporting Ansible 2.8 now (and the maintenance burden). Maybe there's a case for 2.9, maybe. Regardless, I still wish to fix any string based version comparisons.

dw:
> Future Mitogen will only support Ansible 2.10+, so we can remove old checks against old versions of Ansible :)

Don't remove any backwards compatibility code unless it has been thoroughly discussed on a case-by-case basis. Upstream have a horrendous compatibility culture (inherited to a degree from the surrounding ecosystem) that should not leak into this project.

We agreed to conditionally damage backwards compatibility for as long as nobody in the short term exists who is willing to do the work necessary to restore it, but no agreement was made to change compatibility policy for the project as a whole, and certainly no agreement to intentionally remove existing compatibility work.

Separately, the cost of multiple development branches is very much non-trivial for everyone. Starting from installation instructions, OS distributions (Mitogen is packaged in many places), and finally the cost of maintaining multiple distinct versions of the software each with their own testing.

> thinking Ansible 2.8 was EOL a week ago

Regardless of the tragic upstream policies, user playbooks do not suddenly cease to exist because someone decided to externalize costs, nor is it reasonable to require the cost of forced obsolescence to be paid down particularly where it is so easily avoided. It is an abusive bullying tactic, and by adopting upstream policies we enable this abuse.

moreati:
> Don't remove any backwards compatibility code unless it has been thoroughly discussed on a case-by-case basis

In that case, we might need to revert #785.

> Separately, the cost of multiple development branches is very much non-trivial for everyone.

Agreed, I don't enjoy cherry-picking.

PR #788, opened 2021-01-20
tests: Tag resource intensive tasks, to skip when strategy=linear

moreati:
The ansible_mitogen test suite takes over an hour when Ansible is not
accelerated by Mitogen. This change aims to reduce that by skipping
tests with a large number of iterations when the linear strategy is
chosen.

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -30,6 +30,7 @@ jobs:
         MODE: localhost_ansible
         VER: 2.10.0
         STRATEGY: linear
+        ANSIBLE_SKIP_TAGS: resource_intensive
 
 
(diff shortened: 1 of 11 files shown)


moreati:
Mac Vanilla_210_27 job is reduced from 1 hour to approx 15 minutes

s1113950:
image https://dev.azure.com/mitogen-hq/mitogen/_build/results?buildId=32&view=logs&j=c33a2a27-ad1b-5f95-a372-4c6c65381ab9
After removing this, the vanilla Ansible job takes  about 5 min less time than Mitogen 😱 I feel like we should fix that: https://github.com/mitogen-hq/mitogen/issues/804

PR #829, opened 2021-04-20
Skip test_unwriteable when tests are run as root

stefanor:
root can always write to any directory, so this test will fail, incorrectly.


moreati:
> root can always write to any directory

Not strictly true, e.g. read only filesystems, and (I think) the immutable attribute would still prevent root writing. Permission wise, you're right.

stefanor:
Fair enough. I meant in terms of permissions, which is how this test attempts to disable writing.

PR #837, opened 2021-05-14
mitogen.utils.with_router wrapper copies the docstring in addition to the name

### LINKED ISSUE #836: Wrapper do not copy docstring
`mitogen.utils.with_router` does not copy the wrapped functions docstring. Many libraries (e.g. [`click`](https://click.palletsprojects.com)) rely on it though.

```python
from mitogen.utils import with_router

@with_router
def some_click_command(router):
    """This is the description of the command."""
    pass

assert(some_click_command.__doc__ == "This is the description of the command.") # raises exception
```

rqelibari:
Fixes #836.


moreati:
Thank you for doing this, and sorry for the wait. I solved it a slightly different way in #922, noting you as co-author.

PR #844, opened 2021-07-01
Fix issue 827 (collections import)

philfry:
some ansible collections write their own module classes derived from
ansible.module_utils.basic.AnsibleModule, thus lacking that import
in the plugin file. NewStylePlanner was unable to detect these plugins
as Ansiballz.

This PR extends the search pattern for NewStylePlanner to also
detect ansible_collections imports.


moreati:
Thank you for this. Sorry for the long wait.

PR #878, opened 2021-11-18
ci: Fix version comparisons involving double digits

moreati:
See https://gist.github.com/moreati/e7507c5b606b12ec0ddafcb7c8debbf1


moreati:
*sigh* apparently I've been down this road before #787

PR #887, opened 2021-12-20
ensure ssh_args is passed as a list

jamesob:
Passing `ssh_args` as a string creates an implicit (and confusing) error:
```
mitogen.parent.EofError: EOF on stream; last 100 lines received:
ssh: Could not resolve hostname -: Name or service not known
```
Instead, ensure that the type is as expected.


jamesob:
Is this project maintained?

Issue #906, opened 2022-02-23
Mitogen broken by distro 1.7.0


stefanor:
From https://bugs.debian.org/1006330:

Something like issue #590 (see: #610, #622) seems to be back with distro 1.7.0. Presumably because of the change of module layout in https://github.com/python-distro/distro/pull/315

With the Debian packaged ansible 2.10.8, mitogen 0.3.1, on Python 3.9.8. Upgrading from distro 1.6.0 to 1.7.0 on the target broke mitogen:

```
[mux  685121] 14:38:50.354759 D mitogen: PkgutilMethod(): loading 'ansible.module_utils.distro' using <_frozen_importlib_external.SourceFileLoader object at 0x7fa6329f7700> failed: loader for distro cannot handle ansible.module_utils.distro
[mux  685121] 14:38:50.354915 D mitogen: _get_module_via_sys_modules('ansible.module_utils.distro') -> <module 'distro' from '/usr/lib/python3/dist-packages/distro/__init__.py'>
[mux  685121] 14:38:50.355021 D mitogen: sys.modules['ansible.module_utils.distro'].__name__ is incorrect, assuming this is a hacky module alias and ignoring it. Got 'distro', module object: <module 'distro' from '/usr/lib/python3/dist-packages/distro/__init__.py'>
[mux  685121] 14:38:50.355142 D mitogen: ParentEnumerationMethod(): 'ansible.module_utils.distro' is PKG_DIRECTORY: '/usr/lib/python3/dist-packages/ansible/module_utils/distro/__init__.py'
...
[mux  685121] 14:38:51.209954 D mitogen: SysModulesMethod(): sys.modules['ansible.module_utils.distro._distro'] absent or not a regular module
[mux  685121] 14:38:51.210217 D mitogen: ParentEnumerationMethod(): imp.find_module('_distro', [['/usr/lib/python3/dist-packages/distro']]) -> No module named '_distro'
[mux  685121] 14:38:51.210346 D mitogen: get_module_source('ansible.module_utils.distro._distro'): cannot find source
[mux  685121] 14:38:51.210449 D mitogen.responder: could not find source for 'ansible.module_utils.distro._distro'
[mux  685121] 14:38:51.210566 D mitogen.responder: sending ansible.module_utils.distro._distro (0.05 KiB) to ssh.proxy.glacom.eu
[...]

PR #910, opened 2022-03-06
Handle distro 1.7.0's switch from module to package

stefanor:
`ansible.module_utils.distro` replaces itself with either the `distro` module, when it exists, or the bundled distro module.

If a target doesn't have a system `distro` module, it will try to load the bundled distro module (`ansible.module_utils.distro._distro`). This is tricky, because `sys.modules['ansible.module_utils.distro']` points to the
distro package. `ParentEnumerationMethod()` walks up the module tree, to get above this misdirection.

The change from a module to package in distro 1.7.0 broke `ParentEnumerationMethod`'s heuristics. Hardcode avoiding the module `'ansible.module_utils.distro'` as an import starting point.

Fixes: #906

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -31,6 +31,7 @@ v0.3.1.dev0 (unreleased)
 * :gh:issue:`876` `python -c ...` first stage no longer contains tab characters, to reduce size
 * :gh:issue:`878` Continuous Integration tests now correctly perform comparisons of 2 digit versions
 * :gh:issue:`878` Kubectl connector fixed with Ansible 2.10 and above
+* :gh:issue:`906` Support the presence of distro >= 1.7.0.
 
 
 v0.3.0 (2021-11-24)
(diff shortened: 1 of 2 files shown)


willmerae:
Notes for my own understanding
- https://github.com/ansible/ansible/tree/devel/lib/ansible/module_utils/distro contains
  - `__init__.py` which attempts to import https://pypi.org/project/distro/, and on success replaces `sys.modules['ansible.module_utils.distro']` with it.
  - `_distro.py` used as a fallback to populate `sys.modules['ansible.module_utils.distro']`. Bundled versions range from 1.3.0 to 1.6.0 (as of Mar 2022).
- https://pypi.org/project/distro/ <= 1.6.0 ships as a module: `distro.py`
- https://pypi.org/project/distro/ >= 1.7.0 ships as a package: `distro/__init__.py`

moreati:
thanks for the pointer Stefano, you're credited as co-author in the more general fix in #913

PR #933, opened 2022-06-02
Ansible 6 support

moreati:


extmind:
I've tried this branch with Ansible 2.13 and Python 3.10. Its not clear to me if the error is the same as in CI, but here are my findings. While execution only stops later, this is the first traceback that I get:
```
[mux  212055] 17:05:06.914298 D mitogen: While importing 'ansible.module_utils.basic'
  File ".../mitogen/mitogen/master.py", line 1134, in _send_module_and_related
    tup = self._build_tuple(fullname)
  File ".../mitogen/mitogen/master.py", line 1087, in _build_tuple
    for name in self._finder.find_related(fullname)
  File ".../mitogen/mitogen/master.py", line 959, in find_related
    names = self.find_related_imports(name)
  File ".../mitogen/mitogen/master.py", line 933, in find_related_imports
    set(
  File ".../mitogen/mitogen/master.py", line 937, in <genexpr>
    and not is_stdlib_name(name)
  File ".../mitogen/mitogen/master.py", line 140, in is_stdlib_name
    modpath = os.path.abspath(getattr(module, '__file__', ''))
  File "/usr/lib/python3.10/posixpath.py", line 378, in abspath
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```
This is caused by calling `is_stdlib_name` with `jaraco` which has its `__file__` attribute set to `None` (as is explicitly allowed in Python's reference). `jaraco` is installed on my system as a dependency for `setuptools`.
This can be fixed by `modpath = os.path.abspath(getattr(module, '__file__', '') or '')` which would then let the function correctly return `False`, but this may not be adequate in a future situation. Another possibility may be to use a module's `__path__` instead of `__file__` which would be set in `jaraco`'s case, but may also be empty according to the reference.
I think the best solution would be to replace this machinery altogether by `importlib.sys.stdlib_module_names`, but I don't know if this is available for all versions that `mitogen` tries to support. However, it probably should be the preferred variant if available.

PR #942, opened 2022-07-18
Experimental: mitogen: Remove minification of Python source code

moreati:
Done on the basis that sending a few more bytes of the network is quicker than
tokenizing and untokeninzing the modules.


moreati:
As a comparison the slowest jobs in https://dev.azure.com/mitogen-hq/mitogen/_build/results?buildId=382&view=results (revision this branch is based on) took around 24 minutes.

jamesob:
Faster, less code, less complexity... seems like a great change to me.

PR #948, opened 2022-07-28
Provide ansible extra

### LINKED ISSUE #947: Provide ansible as an extras feature on PyPI
I would appreciate if `mitogen` PyPI package has an `ansible` extra with compatible `ansible-core` version bounds. Opting in would mean that automatic update tools such as Dependabot wouldn't suggest updates to `ansible` when `mitogen` is not ready for them.

xfix:
Fixes #947.

  docs/api.rst.


xfix:
Right, that won't work because for 2.10 the package is called `ansible-base`, not `ansible-core`.

xfix:
Eh, I guess I will call it `ansible-core` extra instead, and accept it won't match `ansible-base`. For my use-case of preventing `ansible` updates before mitogen gets updated it's acceptable.

xfix:
Updated the changelog to put this change in "Unreleased" section.

moreati:
If you're till interested, there may be a way this could be work across ansible-base & ansible-core.
1. Release distributions of a new project (e.g. `superluminiferous-aether`) on PyPI with <major.minor> versions that correspond to ansible-base/ansible-core version and that require that corresponding ansible-base/ansible-core <major.minor>
2. Have `mitogen[ansible]` require `superluminiferous-aether >= 2.10, <= 2.16`.

I'm not sure yet, if the benefit outweighs the complexity.

Issue #894, opened 2022-01-15
How to use non-primitive task arguments?


jamesob:
This is more of a question than a bug report, but I'm wondering what the recommended approach is for using rich argument types (e.g. `types.SimpleNamespace`) in tasks.

Due to the whitelisting of unpickleable object types in `Message._find_global`, I get an error similar to the following when trying to pass non-trivial arguments to tasks I'm calling remotely:

```
12:39:54.817173 E mitogen.[ssh.192.168.1.1.sudo.root]: raw pickle was: b'\x80\x02(NX\x08\x00\x00\x00__main__q\x00NX\t\x00\x00\x00provisionq\x01ctypes\nSimpleNamespace\nq\x02)Rq\x03}q\x04X\x08\x00\x00\x00pushoverq\x05h\x02)Rq\x06}q\x07(X\x05\x00\x00\x00tokenq\x08X\x1e\x00\x00\x00auchb1z6xhcpp43gdkifs7ut2s6ffxq\tX\n\x00\x00\x00shell_userq\nX\x1e\x00\x00\x00ud8nou8uccp9wj41fsa91aj7pzqgjgq\x0bubsb\x85q\x0ccmitogen.core\nKwargs\nq\r}q\x0e\x85q\x0fRq\x10tq\x11.'
Traceback (most recent call last):
  File "/home/james/sync/dotfiles/infra/./router.py", line 26, in <module>
    main()
  File "/home/james/sync/dotfiles/infra/./router.py", line 22, in main
    sudo.call(provision, secrets)
  File "/home/james/src/mitogen/mitogen/parent.py", line 2023, in call
    return self.default_call_chain.call(fn, *args, **kwargs)
  File "/home/james/src/mitogen/mitogen/parent.py", line 1980, in call
    return receiver.get().unpickle(throw_dead=False)
  File "/home/james/src/mitogen/mitogen/core.py", line 963, in unpickle
    raise obj
mitogen.core.CallError: mitogen.core.StreamError: cannot unpickle 'types'/'SimpleNamespace'
  File "<stdin>", line 3669, in _dispatch_one
  File "<stdin>", line 3652, in _parse_request
  File "<stdin>", line 952, in unpickle
  File "<stdin>", line 743, in find_class
  File "<stdin>", line 853, in _find_global
```

I understand there are security concerns with unpickling arbitrary objects, but surely datatypes in the standard library don't pose much of a risk?

PR #953, opened 2022-08-11
Allow 

jamesob:
Closes https://github.com/mitogen-hq/mitogen/issues/894.

This change allows end-users to specify allowable globals for unpickling, which is done to allow `CALL_FUNCTION` arguments to be of potentially arbitrary type. I think there's very high convenience here, since it's nice to be able to pass fully-fledged objects to children as call arguments.

This change also adds default deserialization capabilities for `types.SimpleNamespace` (in Python 3), since this is another handy and safe container for task arguments.

If there's conceptual interest here, I can add some tests.


jamesob:
Is this project not maintained?

PR #994, opened 2023-04-30
ansible_mitogen: rediscover python on local connection

qezz:


mistotebe:
I get this issue when delegating in general e.g.:

```yml
hosts: suse-host
  tasks:
    - name: delegated with wrong python interpreter
      command: echo success
      delegate_to: debian-host
```

Will end up with `/usr/bin/python3.6: No such file or directory` and indeed debian doesn't really carry python 3.6 anymore, but ansible decided to pass the discovered interpreted onto it instead of discovering the correct one.

This looks the same as https://github.com/ansible/ansible/issues/61002, maybe mitogen overrides a key part of the fix?

Issue #1053, opened 2024-04-05
Ansible 9 (ansible-core 2.16) support


moreati:
follow on to #1021 

https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_9.html
https://github.com/ansible/ansible/blob/stable-2.16/changelogs/CHANGELOG-v2.16.rst

moreati:
Mitogen 0.3.7 is out https://pypi.org/project/mitogen/0.3.7/ with Ansible 8 and 9 support.

PR #1054, opened 2024-04-05
Ansible 9 support

moreati:
refs #1053


moreati:
Tests pass, but it takes serveral retries. The problem predates adding Ansible 9, but it's getting worse. Causes unknown. Linux+ANsible tests are most susceptible.

Issue #1059, opened 2024-04-09
Tests are slow to run, 15-20 minutes for individual jobs, 40-60 minutes overall


moreati:
Compounds and compounded by #1058

PR #1060, opened 2024-04-10
Speed up test suite

moreati:
refs #1059


moreati:
Reducing the Ansible timeout for all tests reduced the longer jobs by about 2 minutes (20ish to 18ish). This was at the expense of worsening the success rate.

PR #1088, opened 2024-07-16
Remove centos8 CI

jrosser:
Centos-8 is now EOL and the repos are removed, disable the failing CI.


moreati:
Thank you, but removing CentOS 8 tests isn't the way I want to tackling this, at least right now. As you probably saw there are older distros in the mix.

I'll look at fixing the breakage.

Issue #1058, opened 2024-04-09
Tests pass, but it takes serveral retries. The problem predates adding Ansible 9, but it's getting worse. Causes unknown. Linux+Ansible tests are most susceptible.


moreati:
Tests pass, but it takes serveral retries. The problem predates adding Ansible 9, but it's getting worse. Causes unknown. Linux+Ansible tests are most susceptible.

_Originally posted by @moreati in https://github.com/mitogen-hq/mitogen/issues/1054#issuecomment-2040652572_

moreati:
HTTP GET to httpbin.org sometimes times out. Probably best to eliminate the external dependency.

moreati:
https://github.com/mitogen-hq/mitogen/actions/runs/11051662112/job/30702035104?pr=1080 confirms that the Github Actions workflow works with third-party PRs. The Action is able to pull the container images.

moreati:
New contender #1185

moreati:
#1315 occuring more often

moreati:
First sighting https://github.com/mitogen-hq/mitogen/actions/runs/17794076235/job/50577657216?pr=1337
```
 TASK [Create Mitogen test groups name={{ item.name }}] *************************
changed: [localhost] => (item={'name': 'mitogen__group'})
failed: [localhost] (item={'name': 'mitogen__sudo_nopw'}) => changed=false 
  ansible_facts:
    discovered_interpreter_python: /Users/runner/work/mitogen/mitogen/.tox/py313-m_lcl-ans11/bin/python3.13
  ansible_loop_var: item
  item:
    name: mitogen__sudo_nopw
  msg: |-
    Unable to create query for PrimaryGroupID
  name: mitogen__sudo_nopw
```

moreati:
First sighting https://github.com/mitogen-hq/mitogen/actions/runs/17915418248/job/50936370427?pr=683
```
TASK [Start container _raw_params=podman run --name "{{ netconf_container_name }}" --detach --rm --publish "{{ netconf_container_port }}:830" "{{ netconf_container_image }}"] ***
Monday 22 September 2025  12:38:25 +0000 (0:00:00.031)       0:00:55.837 ****** 
task path: /home/runner/work/mitogen/mitogen/tests/ansible/regression/issue_766__get_with_context.yml:34
fatal: [localhost]: FAILED! => 
    changed: true
    cmd:
    - podman
    - run
    - --name
    - sysrepo
    - --detach
    - --rm
[...]

PR #1089, opened 2024-07-26
Fix: termios.error: (22, 'Invalid argument') during `become` on Solaris/Illumos/SmartOS

### LINKED ISSUE #950: become fails on solaris
When using with recent SmartOS (Solaris/Illumos), mitogen works as expected (and super fast) as long as I don't use become. If I use become, this error occurs in tcgetattr on the master connection; if I avoid calling that for the master, that problem goes away (obviously), but mitogen times out waiting for MITO000 prompt.

I'm relatively new to mitogen, very experienced in Ansible and Python (I've submitted upstream patches to ansible and cpython) and can provide a test environment if necessary. If you can point me to some diagnostics or other thoughts, I'm interested in getting this to work. It's seemingly so close, but I'm not sure of the differences between this working without sudo and not working with sudo.

* Which version of Ansible are you running?
2.12.5

* Is your version of Ansible patched in any way?
no

* Are you running with any custom modules, or `module_utils` loaded?
no

* Have you tried the latest master version from Git?
yes
* Do you have some idea of what the underlying problem may be? 
  https://mitogen.networkgenomics.com/ansible_detailed.html#common-problems has
  instructions to help figure out the likely cause and how to gather relevant
  logs.

looks to be a difference in pty handling after openpty. The first issue was a call (in disable_echo) to:
    old = termios.tcgetattr(fd)
which for the fd in question (master) raises  termios.error: (22, 'Invalid argument')

If I put a try:except: around the call to disable_echo for master, the command times out waiting for the MITO000 prompt. 

```
    master_fp = os.fdopen(master_fd, 'r+b', 0)
    slave_fp = os.fdopen(slave_fd, 'r+b', 0)
    try:
      disable_echo(master_fd)
    except:
      pass
    disable_echo(slave_fd)
```

* Mention your host and target OS and versions
SmartOS 21.4 (most recent patches) on both sides.

* Mention your host and target Python versions
Python 3.9.13 on both

* If reporting a crash or hang in Ansible, please rerun with -vvv and include
  200 lines of output around the point of the error, along with a full copy of
  any traceback or error text in the log. Beware "-vvv" may include secret
  data! Edit as necessary before posting.

Attached as pre_patch and post_patch (with the try:except around `disable_echo(fd_master)`)

* If reporting any kind of problem with Ansible, please include the Ansible
  version along with output of "ansible-config dump --only-changed".

```
DEFAULT_ROLES_PATH(/root/ansible-web/ansible.cfg) = ['/root/ansible']
DEFAULT_STRATEGY(/root/ansible-web/ansible.cfg) = mitogen_linear
DEFAULT_STRATEGY_PLUGIN_PATH(/root/ansible-web/ansible.cfg) = ['/root/mitogen/ansible_mitogen/plugins/strategy']
DEFAULT_VAULT_PASSWORD_FILE(env: ANSIBLE_VAULT_PASSWORD_FILE) = /root/.vault_password
HOST_KEY_CHECKING(/root/ansible-web/ansible.cfg) = False
NETCONF_SSH_CONFIG(env: ANSIBLE_NETCONF_SSH_CONFIG) = True
```
[pre_patch.txt](https://github.com/mitogen-hq/mitogen/files/9274751/pre_patch.txt)
[post_patch.txt](https://github.com/mitogen-hq/mitogen/files/9274753/post_patch.txt)

gaige:
This fixes compatibility with Solaris/Illumos/SmartOS, addressing an issue that shows up most frequently with become. The issue was mostly due to differences in how the TTY driver is handled and the pty driver not supporting echo on both sides of the pipe (as designed, from a Solaris point of view).

Fixes #950


oetiker:
hmmm I tried the patch and it did not suffice to make things work ... what I did find was that probably the 

```
ERROR! [mux  1590247] 11:29:26.373979 E mitogen.[ssh.micd-omnios-adm]: ExternalContext.main() crashed
  File "<stdin>", line 4196, in main
  File "<stdin>", line 3933, in run
  File "<stdin>", line 679, in _profile_hook
  File "<stdin>", line 3916, in _dispatch_calls
  File "<stdin>", line 1237, in __iter__
  File "<stdin>", line 1223, in get
  File "<stdin>", line 2843, in get
  File "<stdin>", line 2807, in _make_cookie
struct.error: int too large to convert
```

I also tried a patch to make cfmakeraw happier:

```python
def cfmakeraw(tflags):
    """
    Given a list returned by :py:func:`termios.tcgetattr`, return a list
    modified in a manner similar to the `cfmakeraw()` C library function, but
    additionally disabling local echo.
    """
    # BSD: github.com/freebsd/freebsd/blob/master/lib/libc/gen/termios.c#L162
    # Linux: github.com/lattera/glibc/blob/master/termios/cfmakeraw.c#L20
    iflag, oflag, cflag, lflag, ispeed, ospeed, cc = tflags
    if IS_SOLARIS:
        iflag &= ~flags('IGNBRK BRKINT PARMRK ISTRIP INLCR IGNCR ICRNL IXON')
        oflag &= ~flags('OPOST')
        lflag &= ~flags('ECHO ECHONL ICANON ISIG IEXTEN')
        cflag &= ~flags('CSIZE PARENB')
        cflag |= flags('CS8')
    else:
        iflag &= ~flags('IMAXBEL IXOFF INPCK BRKINT PARMRK '
                        'ISTRIP INLCR ICRNL IXON IGNPAR')
        iflag &= ~flags('IGNBRK BRKINT PARMRK')
        oflag &= ~flags('OPOST')
        lflag &= ~flags('ECHO ECHOE ECHOK ECHONL ICANON ISIG '
                    'IEXTEN NOFLSH TOSTOP PENDIN')
        cflag &= ~flags('CSIZE PARENB')
        cflag |= flags('CS8 CREAD')
    return [iflag, oflag, cflag, lflag, ispeed, ospeed, cc]
```

but because of the cookie problem I did not see whether this helped any

gaige:
@oetiker  Interesting failure mode. What's the context of the test you were doing where `make_cookie` failed?  I was running Ansible 2.14.5 on my test machine on very recent SmartOS going against a series of similar-version SmartOS hosts.

gaige:
I reran my tests on Ansible 2.15.12 and was seeing the same success. Although maybe you are seeing a completely different problem that I was here.

oetiker:
yes I am pretty sure my problem is a different one ... maybe related to a different range for process ids on solaris or some such ?

gaige:
> yes I am pretty sure my problem is a different one ... maybe related to a different range for process ids on solaris or some such ?

What Solaris or derivative are you running? We're running SmartOS and I'm not seeing crashes in `_make_cookie`, however since it's basically a `pack` call, differing sizes would definitely be suspicious.

moreati:
Notes to myself, 
- OpenSolaris: initial Open Source release of Solaris by Sun, discontinued by Oracle
- Solaris Express, Oracle Solaris: proprietary releases
- Illumos: Foundation/umbrella continuing OpenIndiana (fka OpenSolaris), basis of multiple "Illumos distributions"
- SmartOS: Illumos distribution that incorporates hypervisors & cloud platform
- Helios: Illumos distribution by Oxide Computers (WIP)

Other `IS_*` constants in Mitogen
```
➜  mitogen git:(master) ✗ ag -s '(\b|\.)IS_[A-Z0-9_]+ +='
mitogen/parent.py
149:IS_LINUX = os.uname()[0] == 'Linux'

mitogen/core.py
170:IS_DEAD = 999
203:    IS_WSL = 'Microsoft' in fp.read()
206:    IS_WSL = False
```

Questions to ponder (answers are not required to merge this)
1. Could an Illumos distro be added as a test target in CI (e.g. VM, container voodoo)? Externally hosted?
2. Is it worth creating another top level `IS_*` constant? Is it worth keeping the existing ones?
3. Can `disable_echo()` be unit tested? Should it?

mitogen/parent.py
```diff
@@ -148,6 +148,9 @@ def _ioctl_cast(n):
 
 IS_LINUX = os.uname()[0] == 'Linux'
 
+IS_SOLARIS = os.uname()[0] == 'SunOS'
```
moreati, line 151: How much can we rely on different Illumos distributions sticking to 'SunOS'?

mitogen/parent.py
```diff
@@ -479,7 +482,10 @@ def openpty():
 
     master_fp = os.fdopen(master_fd, 'r+b', 0)
     slave_fp = os.fdopen(slave_fd, 'r+b', 0)
-    disable_echo(master_fd)
+    try:
+      disable_echo(master_fd)
+    except:
```
moreati, line 487: I won't accept a bare `except:` or `except Exception:`.

1. What specific failure(s) are you trying to catch?
2. Is ignoring the exception an acceptable fallback in the general case?

mitogen/parent.py
gaige: In that case, because it's the "base" (if you will) OS, and there's no real hierarchy in `uname` all the derived versions have kept it for compatibility.  It's more of a platform specifier at this point.

mitogen/parent.py
gaige: I can make that Solaris-specific. The failure is definitely a documented behavior on the solaris platforms.

gaige:
@moreati : not sure what is causing these CI failures, but they don't seem directly related to (or unique to) this patch.

moreati:
> not sure what is causing these CI failures, but they don't seem directly related to (or unique to) this patch.

They're due to #1058, almost certainly unrelated to your change. To re-run tests for a PR add a comment containing

>    /AzurePipelines run

gaige:
@moreati Any other requests for this one now that it's passing checks?

moreati:
> @moreati Any other requests for this one now that it's passing checks?

Yes. Two suggested changes are still open (see previous comments), and I want to answer "what are the consequences of leaving echo enabled?" before merging.

gaige:
> > @moreati Any other requests for this one now that it's passing checks?
> 
> Yes. Two suggested changes are still open (see previous comments), and I want to answer "what are the consequences of leaving echo enabled?" before merging.

The reason I asked is that I looked through the comments here and *believe* that I have answered the questions or have addressed them in code. If you can give me a reference to the outstanding suggested changes, I'll gladly address them. Sorry if I'm being dense here.

The question of "what are the consequences of leaving echo enabled?" isn't quite the question for Solaris. In the case of mitogen, you're disabling echo on both sides of the pty, which works in linux. For Solaris, the pty is streams-based and because of the way it works, the  `termio` ioctls are supported by the subsidiary (slave) device on Solaris. Setting termios attributes, such as ECHO on Solaris (and other Unix98 systems) is not allowed and will result in the invalid parameter error.

moreati:
> The reason I asked is that I looked through the comments here and _believe_ that I have answered the questions or have addressed them in code. If you can give me a reference to the outstanding suggested changes

You should be able to see the suggested changes on this page, and accept them using the buttons below each. This is how they look to me

gaige:
> You should be able to see the suggested changes on this page, and accept them using the buttons below each. This is how they look to me

Oddly enough, I don't see that on my screen, and searching for those words on the page doesn't find them either.  With that said, let me apply the patches.

mitogen/parent.py
```diff
@@ -479,7 +482,10 @@ def openpty():
...
+      pass
     disable_echo(slave_fd)
```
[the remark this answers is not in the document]
moreati: It is probably the call to `termios.tcgetattr(master_fd)` that is raising `termios.error` (i.e. we never even reach `termios.tcsetattr()`)

```python
import os, termios

master, slave = os.openpty()
flags = termios.tcgetattr(slave)
print(flags[0] & termios.IGNBRK)
os.close(slave)
os.close(master)
```
```
alex@oiv2404:~$ python termios_play.py
  File "/export/home/alex/termios_play.py", line 4, in <module>
    flags = termios.tcgetattr(master)
termios.error: (22, 'Invalid argument')
```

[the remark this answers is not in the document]
moreati: 🤦 that's plainly stated in the description of #950.

mitogen/parent.py
```diff
@@ -479,7 +482,8 @@ def openpty():
 
     master_fp = os.fdopen(master_fd, 'r+b', 0)
     slave_fp = os.fdopen(slave_fd, 'r+b', 0)
-    disable_echo(master_fd)
+    if not IS_SOLARIS:
+      disable_echo(master_fd)
```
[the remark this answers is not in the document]
moreati: This question is probably based on a false premise - that these are independant attributes of each end (master and slave). I'm starting to think they're attributes of the channel. Calling `disable_echo()` twice may be redundant on Linux as well.

moreati (approved):
I still can't find primary documentation that clearly states Solaris-likes only accept the slave fd for `tcgetattr()` and `tcsetattr()`, but the exerimental evidence has me satisfied.

Issue #1073, opened 2024-05-14
Python 3.13 support


moreati:

moreati:
Affected by https://github.com/python/cpython/issues/115911, may be a blocker or require caveats on functionality

PR #1146, opened 2024-10-07
Remove get_password_hash, unused

stefanor:
spwd is removed in Python 3.13. But fortunately, this function itself is never used.

Part of: #1073

### DIFF
--- a/examples/the_basics.py
+++ b/examples/the_basics.py
@@ -10,7 +10,6 @@ from __future__ import print_function
 import hashlib
 import io
 import os
-import spwd
 
 import mitogen.core
 import mitogen.master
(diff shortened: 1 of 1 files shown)


moreati:
refs 1073

moreati:
Notes
- `spwd.getspnam(name)` requires root (or *mumble* POSIX capability). It reads `/etc/shadow`. There's no interaction with any stdio or a tty
- `getpass.getpass(prompt='Password: ', stream=None)` requires no additional privilege. It writes to `stream` if supplied, or the controlling terminal if available, or stderr otherwise. It reads from `stream` if provided (I think), or the controlling tty, or stdin (with a warning).

moreati:
> This is just example code that is never called by anything.

Doubly so, get_password() (prev. `get_password_hash()`) isn't even called within its own example script.

moreati (changes requested):
I think I prefer deleting the entire function; on the basis the function is unused and this file advertises itself as just the basics. Do you have a preference, or know of a reason to keep it?

stefanor:
Let's delete it. I was assuming it was there as a demonstrative example.

moreati (approved):

Issue #1116, opened 2024-09-06
feat(ansible): Handle templated `ssh_user`


setpill:
ansible-mitogen does not handle templated values in `ssh_user` .

Original context from !956 :

> [..] this does fix the `ssh_common_args`, but I am running also into the issue that a templated `ssh_user` is not correctly handled.

_Originally posted by @setpill in https://github.com/mitogen-hq/mitogen/issues/956#issuecomment-2225208105_

setpill:
Usecase: in our environment we have ansible roles that have tasks that need to be executed by different users. For legacy and security reasons we cannot assume these users to have fixed names, so we need to provide them in variable form.

Let's say for example that a role executes some tasks as `build_user` and some tasks as `deploy_user`. We would then have a task (block) in the form of:

```yml
- name: Build something
  vars:
    ansible_user: "{{ build_user }}"
    ansible_password: "{{ build_pass }}"
    ansible_sudo_pass: "{{ build_pass }}"
  block:
    - [..]
 ```

Ansible-mitogen is not currently a drop-in replacement for this scenario, as it will not template these variables and pass the literal string `{{ build_user }}` as the SSH user, which fails.

moreati:
> Ansible-mitogen is not currently a drop-in replacement for this scenario, as it will not template these variables and pass the literal string `{{ build_user }}` as the SSH user, which fails.

I'm investigating  2 Mitogen bugs that are probably affecting your playbook
1. Many of the `ansible_*` (connection) variables don't get templated (varies issues already filed)
2. For the few that are templated, anything in a play's `vars:` block is ignored.

setpill:
Thanks for your great work! Mitogen really is a much faster engine for ansible, would be awesome to get it closer to that "drop-in replacement" ideal :)

Issue #1022, opened 2023-09-10
ansible_user and ansible_ssh_host variables not evaluated by mitogen


vdanjean:
* Which version of Ansible are you running?
  7.7.0+dfsg-1 from debian unstable

* Is your version of Ansible patched in any way?
  no

* Are you running with any custom modules, or `module_utils` loaded?
  no

* Have you tried the latest master version from Git?
  no

* Do you have some idea of what the underlying problem may be?
  It seems that ansible variables related to ssh (at least) are not evaluated

* Mention your host and target OS and versions
  host is Debian unstable, target is Debian stable (bookworm)

* Mention your host and target Python versions
  host: 3.11.4-5+b1
  target: 3.11.2-1+b1

I reproduce the problem with a small inventory and task file:
ansible_user and ansible_ssh_host are defined to a non-constant in the inventory file (group `dj`):
```
---
dj:
  hosts:
    aya:
  vars:
    setup_ansible_user_login: vdanjean-ansible-client

    local_ssh_domain: ""
    ansible_ssh_host: "{{ inventory_hostname }}{{ local_ssh_domain }}"

    ansible_user: "{{ setup_ansible_user_login }}"
```
My test playbook is the following
```
---

 - name: tests ssh
   hosts: all
   gather_facts: false
   tasks:
     - block:
         - set_fact:
             foo: bar
             #ansible_ssh_host: "{{ ansible_ssh_host }}"
             #ansible_user: "{{ ansible_user }}"
         - command: hostname
           register: res
           ignore_unreachable: true
       always:
         - debug: var=res
```
As it is, running it with default stategy works. But, with `strategy = mitogen_linear`, I got:
```
$ ansible-playbook -vvv -l aya test.yml 
ansible-playbook [core 2.14.9]
  config file = /home/vdanjean/travail/perso/Claire/NGS/gitolite/vdanjean-chu-ansible/bug/ansible.cfg
  configured module search path = ['/home/vdanjean/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/vdanjean/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible-playbook
  python version = 3.11.5 (main, Aug 29 2023, 15:31:31) [GCC 13.2.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
[...]

PR #1150, opened 2024-10-08
tests: Use templated ansible_user for localhost

moreati:
refs #1022, #1116


moreati:
Turns out the only connection parameter of `ansible.plugins.connection.local` is connection pipelining, for which Mitogen does its own thing.

Issue #1040, opened 2024-02-23
The attribute 'remote_user' of the object 'connection._play_context' contains the variable name instead of its value.


macksyma:
I've discovered the incorrect transmission of the ```remote_user``` variable's value when using ```delegate_to``` with Mitogen 3.4. For example, let's take a simple task:

```yml
    - name: "test_task"
      shell: "ls -lah"
      delegate_to: "{{ test_jump_box }}"
      remote_user: "{{ jumper_test_user }}"
```

Here is the error I've received executing this task:

```
fatal: [localhost -> XXXXXXXXXX]: UNREACHABLE! => {
    "changed": false,
    "msg": "EOF on stream; last 100 lines received:\nremote username contains invalid characters\r",
    "unreachable": true
}
```

In the debug log, I see that the ssh command line contains the name of the variable transmitted into the ```remote_user``` - ```{{ jumper_test_user }}```, instead of the actual value of this variable:

```
[mux  2197866] 12:52:15.298866 D mitogen.parent: command line for Connection(None): ssh -o "LogLevel ERROR" -l "{{ jumper_test_user }}" -o "Compression yes" -o "ServerAliveInterval 30" -o "ServerAliveCountMax 10" -o "BatchMode yes" -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -o "GlobalKnownHostsFile /dev/null" -F ssh.cfg ...
```

After further investigation, I've discovered that Mitogen takes the username for ssh from object ```_connection._play_context```, transmitted to Mitogen's ```ActionModuleMixin``` from Ansible's ```TaskExecutor```. Inside the ```ActionModuleMixin``` object, I've found that ```_self._play_context.remote_user``` contains the correct value of the variable ```jumper_test_user```, but the connection object ```self._connection._play_context.remote_user``` contains the variable name, including brackets, just like I saw it in the ssh command line.

I've checked how the method ```_execute```of Ansible's class ```TaskExecutor``` updates attributes of its own ```self._play_context``` and ```self._connection._play_context```, and found that at some point, ```self._play_context``` of ```TaskExecutor``` also contains the name of variables instead of their value. But then, it executes the method ```post_validate()``` for its ```self._play_context``` (PlayContext inherits it from its parent classes):

```python
self._play_context.post_validate(templar=templar)
```

[...]

PR #1153, opened 2024-10-09
tests: Templated "remote_user" provided as Ansible playbook keyword

moreati:
The password is provided as a variable because there is no corresponding keyword. I get the impression that keywords are considered a legacy mechanism, so most (new) options are only overridable by variables.

The port is proved as a variable for now, to test remote_name in isolation.

refs #1040

### DIFF
--- a/tests/ansible/integration/ssh/all.yml
+++ b/tests/ansible/integration/ssh/all.yml
@@ -3,5 +3,6 @@
 - import_playbook: password.yml
 - import_playbook: timeouts.yml
 - import_playbook: templated_by_inv.yml
+- import_playbook: templated_by_play_keyword.yml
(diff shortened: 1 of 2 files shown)


moreati:
```
PLAY [integration/ssh/templated_by_play_keyword.yml] ***************************

TASK [meta _raw_params=reset_connection] ***************************************
Wednesday 09 October 2024  12:32:10 +0000 (0:00:00.314)       0:01:43.953 ***** 

TASK [Templated variables in play keywords ] ***********************************
Wednesday 09 October 2024  12:32:10 +0000 (0:00:00.022)       0:01:43.975 ***** 
ok: [tt-bare]
```

Issue #1083, opened 2024-05-27
Jinja expressions are not evaluated in (not only) credential variables


mordekasg:
### OS / Environment
Debian 11

### Ansible version
```
ansible [core 2.14.6]
  config file = /home/test/host/ansible.bs/ansible.cfg
  configured module search path = ['/home/test/host/ansible.bs/local_modules']
  ansible python module location = /home/test/env.ansible.bs/lib/python3.9/site-packages/ansible
  ansible collection location = /home/test/host/ansible.bs/collections:/usr/share/ansible/collections
  executable location = /home/test/env.ansible.bs/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110] (/home/test/env.ansible.bs/bin/python)
  jinja version = 3.0.1
  libyaml = True
```

### Environment variables for strategy plugin
```
ANSIBLE_STRATEGY=mitogen_linear
ANSIBLE_STRATEGY_PLUGINS=/home/test/env.ansible.bs/src/mitogen/ansible_mitogen/plugins/strategy
```

### Mitogen version
```
master branch
```

### Problem
Jinja expressions are not evaluated by mitogen plugin

#### Description
Ansible host: **srv1** has credentials configured using Jinja expressions from vault variables.
Ansible host: **srv2** has credentials configured directly as host variables.

#### Vault file
Path: **vault.test.yml**
```
ANSIBLE_LOCALHOST_SRVS_USER: "root"
ANSIBLE_LOCALHOST_SRVS_PASSWORD: "okmqwe123"
```
#### Inventory file
Path: **inventory/local/local.yml**
```
all:
  children:
    test_srvs:
      hosts:
        srv1:
          ansible_host: 127.0.0.1
          ansible_user: "{{ ANSIBLE_LOCALHOST_SRVS_USER }}"
          ansible_password: "{{ ANSIBLE_LOCALHOST_SRVS_PASSWORD }}"
          ansible_ssh_user: "{{ ansible_user }}"
          ansible_ssh_password: "{{ ansible_password }}"
        srv2:
          ansible_host: 127.0.0.1
          ansible_user: root
          ansible_password: okmqwe123
          ansible_ssh_user: root
          ansible_ssh_password: okmqwe123
```

#### Command
```
ansible -e @vault.test.yml -i inventory/local/local.yml -m ping test_srvs
```
#### Output
```
srv2 | SUCCESS => {
    "changed": false,
[...]

Issue #1133, opened 2024-10-06
ci: Migrate macOS runners from macOS 12


moreati:
The macOS 12 action runner is being removed, brownouts in Nov, death on 2024-12-04 https://github.com/actions/runner-images/issues/10721

moreati:
Affects Azure DevOps as well  as GitHub Actions

moreati:
https://github.com/mitogen-hq/mitogen/actions/runs/11669080981/job/32490230504?pr=1176
```
TASK [Create /tmp/512mb.zero creates=/tmp/512mb.zero, _raw_params=dd if=/dev/zero of=/tmp/512mb.zero bs=1048576 count=512;
chmod go= /tmp/512mb.zero
] ***
Monday 04 November 2024  17:08:28 +0000 (0:00:00.007)       0:00:26.162 ******* 
changed: [target]
TASK [Fetch /tmp/512mb.zero src=/tmp/512mb.zero, dest=/tmp/fetch-{{ inventory_hostname }}-512mb.zero, flat=True] ***
Monday 04 November 2024  17:08:30 +0000 (0:00:01.338)       0:00:27.501 ******* 
ERROR! [mux  5271] 17:08:37.240824 E mitogen: <Stream unix_client.5683 #1dd0> crashed
Traceback (most recent call last):
  File "/Users/runner/work/mitogen/mitogen/mitogen/core.py", line 3677, in _call
    func(self)
    ~~~~^^^^^^
  File "/Users/runner/work/mitogen/mitogen/mitogen/core.py", line 1923, in on_transmit
    self.protocol.on_transmit(broker)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/runner/work/mitogen/mitogen/mitogen/core.py", line 2371, in on_transmit
    self._writer.on_transmit(broker)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/runner/work/mitogen/mitogen/mitogen/core.py", line 2111, in on_transmit
    written = self._protocol.stream.transmit_side.write(buf)
  File "/Users/runner/work/mitogen/mitogen/mitogen/core.py", line 2237, in write
    written, disconnected = io_op(os.write, self.fd, s)
                            ~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/mitogen/mitogen/mitogen/core.py", line 595, in io_op
    return func(*args), None
[...]

PR #1175, opened 2024-10-29
ansible_mitogen: Templated ssh executable

moreati:
Adding a the tt-ssh-executable test target uncovered Ansible bug https://github.com/ansible/ansible/issues/84238 during `meta: reset_connection` tasks. So this commit includes a workaround for affected versions of Ansible.

refs #1083,


moreati:
Awaiting end of today's GitHub Actions macOS 12 (#1133) brownout to complete tests and merge.

PR #1176, opened 2024-11-04
CI: Migrate to from macOS 12 to 13 test runners

moreati:
refs #1133


moreati:
macOS 13 images no longer have CPython 2.7. PyPy 2.7 is there, and might be worth exploring in the future. For now I'm dropping the macOS Python 2.7 jobs. Linux CI jobs will have to carry the Python 2.x testing torch

PR #1190, opened 2024-11-07
Prepare v0.3.17 and some washup tests

moreati:


moreati:
Removing the port by keywords test. Vanilla Ansible does not template values provided by keyword. I think this is the second time I've disocvered this.

```
PLAY [integration/ssh/templated_by_play_keyword.yml] ***************************

TASK [meta _raw_params=reset_connection] ***************************************
Thursday 07 November 2024  12:34:39 +0000 (0:00:00.480)       0:01:48.836 ***** 
ERROR! Invalid type for configuration option plugin_type: connection plugin: ssh setting: port (from keyword: port): invalid literal for int() with base 10: "{{ hostvars[groups['test-targets'][0]].ansible_port | default(22) }}"
```

PR #1217, opened 2025-01-11
ansible_mitogen: Add all ansible_freeipa modules to the always-fork list.

### LINKED ISSUE #1216: ansible-freeipa modules require forking
Most ansible-freeipa modules initialize a connection to the FreeIPA API that persists for the duration of a single module's execution. ipalib, where this method lives, assumes that only a single bootstrapped API exists at any one time, and that any future invocations of bootstrap() are preceded by a finalize() call.

While this could in theory be fixed in freeipa, it's a relatively invasive fix which may not be the best option. As a stopgap, the relevant modules should be added to Mitogen's internal always-fork list.

Nihlus:
This PR adds all of ansible_freeipa's modules to the always-fork list due to assumptions about interpreter isolation made in the upstream codebase.

Fixes #1216.


moreati (changes requested):
Looks good, but needs a https://github.com/mitogen-hq/mitogen/blob/master/docs/changelog.rst entry

Nihlus:
Fixed, thanks :)

moreati (approved):

Issue #711, opened 2020-04-21
Mitogen is not compatible with 'become_exe = "sudo su -" ' in ansible.cfg


tdluckyboy:
Ansible 2.9.2 with ansible.cfg:
  [defaults]
  inventory = /home/xxx/ansible/staging
  remote_user = xxx
  host_key_checking = False
  callback_whitelist = profile_tasks
  strategy_plugins = /home/xxx/ansible/mitogen-0.2.9/ansible_mitogen/plugins/strategy
  strategy = mitogen_linear

  [privilege_escalation]
  become = True
  become_method = su
  become_user = root
  become_exe = "/usr/bin/sudo su -"

Mitogen 0.2.9
Python (host  2.7.5 and target is 2.6.6)
OS(Host is CentOS 7.7 and target is RHEL 6.10)

Error:
ansible-playbook site.yml  -vvv

[mux  18510] 19:00:25.388312 D mitogen.importer.[ssh.xxxxxx]: received mitogen.su
[mux  18510] 19:00:25.390934 D mitogen.parent.[ssh.xxxxxx]: creating connection to context 4 using mitogen.su
[mux  18510] 19:00:25.399658 D mitogen.importer.[ssh.xxxxxx]: encodings.ascii is submodule of a locally loaded package
[mux  18510] 19:00:25.400047 D mitogen.importer.[ssh.xxxxxx]: encodings.base64_codec is submodule of a locally loaded package
[mux  18510] 19:00:25.400244 D mitogen.importer.[ssh.xxxxxx]: encodings.base64 is submodule of a locally loaded package
[...]

PR #1243, opened 2025-02-05
Experimental: pass first stage parameters in argv

moreati:


moreati:
The old arrangemet was 798 bytes, as measured by preamble_size.py, compression tweaks and code golfing dropped that to approx 786 bytes. The new arrangement is 822 bytes, including compression tweaks and golfing.

Benefits of the new arrangement are
- static first stage
  - transparency - the scary base64 blob can be reversed once, analysed once, subjected to much tighter filtering or allow listing
  - less magic - no 'constants' that are actually string labels replaced at runtime
  - looser coupling - between the `python -c ...` stage (zeroth stage) and the first stage.

Potential benefit
  - speedup - the first stage doesn't have to be regenerated for each context/connection.

moreati:
refs #711

moreati:
`env -S ...` might be of use
```
-S, --split-string=S
    process and split S into separate arguments; used to pass multiple arguments on shebang lines
```
-- https://manpages.debian.org/trixie/coreutils/env.1.en.html

PR #1255, opened 2025-02-28
Use regex matching for modules that should always fork

Nihlus:
The current logic used to match module names that should always fork has room for improvement. As exemplified by the recent addition of the FreeIPA modules, which was a rather long list, the ability to use patterns instead of exact string matches would improve readability and long-term maintainability of the mechanism.

This PR changes the module set into a set of regular expression strings and compiles them into a single joint regex that we then use to match against the module name.

### DIFF
--- a/ansible_mitogen/planner.py
+++ b/ansible_mitogen/planner.py
@@ -341,71 +341,14 @@ class NewStylePlanner(ScriptPlanner):
     #: to some terminal leakage that cannot be worked around in any sane
     #: manner.
     ALWAYS_FORK_MODULES = frozenset([
-        'dnf',  # issue #280; py-dnf/hawkey need therapy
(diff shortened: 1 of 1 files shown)


moreati:
Notes/random thoughts
- Agree the long `frozenset` is ugly and too brittle
- If we're changing the interface/meaning of the attribute ALWAYS_FORK_MODULES, then it's probably not worth keeping.
- Perhaps hide implementation details behind a method like `_module_requires_fork(self, modname)`
- Could it become runtime configurable? E.g. Ansible plugin option, environment variable

Sorry for the long wait to look at this. I plan to give it more attention in the next few weeks

moreati:
Note to self: Possible inspiration for user configurable ALWAYS_FORK_MODULES through an environment variable https://github.com/python/cpython/pull/138149

Issue #1258, opened 2025-04-20
Support Ansible 12 (ansible-core 2.19)


stefanor:
Debian is looking at Ansible 12 support early, as we're entering the early stages of freeze now for Debian 13 (Trixie) and expect Ansible 12 to release well before we freeze solid.

carnil:
FWIW, the report in Ansible is at: https://github.com/ansible/ansible/issues/85017

upekkha:
Ansible 12 is deprecating custom strategy plugins like mitogen: ansible/ansible/issues/84725

>  Use of strategy plugins not included in ansible.builtin are deprecated and do not carry any backwards compatibility guarantees. No alternative for third party strategy plugins is currently planned. This feature will be removed in a future release.

moreati:
Notes to self

- Debian 13 (Trixie)
  - 15 Apr 2025 soft freeze, 15 May hard freeze 
  - Ships Python 3.13
- Ansible 12 (core 2.19)
  - 16 Apr 2025 core feature freeze, 26 May core RC1 
  - Controller: Python 3.11 - 3.13
  - Targets: Python 3.8 - 3.13

Refs
1. https://wiki.debian.org/DebianTrixie
2. https://distrowatch.com/table.php?distribution=debian
3. https://docs.ansible.com/ansible/devel/roadmap/COLLECTIONS_12.html
4. https://docs.ansible.com/ansible/devel/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix

moreati:
`ansible.parsing.utils.jsonify` was removed in https://github.com/ansible/ansible/pull/84621 (Templating overhaul, Data Tagging), some fallout reported/discussed in https://github.com/ansible/ansible/issues/85017, restored with deprecation warning in https://github.com/ansible/ansible/pull/85032.

moreati:
Prior to removal/restoration`ansible.parsing.utils.jsonify.jsonify()` changed very little. Differences to `json.dumps()`

1. `jsonify(None)` -> `u'{}'` vs `dumps(None)` -> `u'null'`
2. `jsonify(..., format: bool)` vs `dumps(..., indent: int|None, ...)`
3. `jsonify()` always applies `sort_keys=True`.
4. `jsonify()` prefers `ensure_ascii=False`, falling back to `ensure_ascii=True` on `UnicodeDecodeError`. I'm unsure what would trigger that exception only in the former case.

Refs
[...]

PR #1264, opened 2025-04-29
Ansible 2.19 (review/sandbox)

moreati:
DO NOT MERGE

Sandbox for collaborating on #1258 by @stefanor


stefanor:
5a05332 should be dropped, upstream's revert has 

```
    if result is None:
        return "{}"
```

That's an issue I've run into in my own use with this patch.

moreati:
Progress update: think I'm over the hump, wrt to getting the Mitogen test suite passing with ansible-core 2.19. Currently debugging async tasks and the test https://github.com/mitogen-hq/mitogen/blob/master/tests/ansible/integration/async/multiple_items_loop.yml. The simpler reproduction is 

```yaml
- hosts: localhost
  gather_facts: false
  tasks:
    - shell: sleep 2; echo hi
      # Wait upto <async> seconds for completion
      async: 5
      # If poll > 0 then check for completion every <poll> seconds
      # If poll == 0 then proceeed to next task immediately
      poll: 1
      register: job
      changed_when: false
```

It works with https://github.com/ansible/ansible/tree/6fc592df9b81efed969b9950c9fae373e2574a6a (just before datatags PR https://github.com/ansible/ansible/pull/84621), but not later revisions.

moreati:
Async tasks are timing out because the Ansible Modules dynamically import Python modules in the package `ansible.module_utils._internal._json._profiles`, and with an `async` task with mitogen this happens after the subtree/econtext has been detached from the master. Hence any attempt to import a module (that would normally be fetched on the wire, and delivered from the master) fails (timeout?).

Adding import statements to the Ansible modules (e.g. `command`), and/or `ansible.module_utils.basic` means the modules are requested before the econtext disconnects e.g.

```python
from ansible.module_utils._internal._json._profiles import _module_legacy_c2m
from ansible.module_utils._internal._json._profiles import _module_legacy_m2c
from ansible.module_utils._internal._json._profiles import _module_modern_c2m
from ansible.module_utils._internal._json._profiles import _module_legacy_m2c
```

This is a hacky debugging aid, not a solution. Solutions might include
- preload the modules another way
- detach partially, e.g. leaving the import plumbing attached

moreati:
Split core mitogen changes into their own PR/release, and rebased Ansible 12 changes on that

PR #1277, opened 2025-06-08
docs: Fix website download link when there is a pre-release

### LINKED ISSUE #1276: docs: Website installation instructions incorrectly treat a pre-release as stable
Because the `changelog_version()` regex only looks at the start of each line it found "0.3.25" in https://github.com/mitogen-hq/mitogen/blob/837111c6d58d38b4a486316118120477d001d950/docs/changelog.rst#L26-L27 whereas the intended behaviour was to match only release versions. So https://mitogen.networkgenomics.com/ansible_detailed.html#installation says

> 2. Download and extract [mitogen-0.3.25.tar.gz](https://files.pythonhosted.org/packages/source/m/mitogen/mitogen-0.3.25.tar.gz).

even though that release doesn't exist.

Inverse of #1157. Unintended knock on of #1211

moreati:
The previous regex was incorrectly matching a prefix (e.g. 1.2.3) of a pre-release (e.g. 1.2.3a1, 1.2.3rc1).

fixes #1276

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,8 +7,9 @@ sys.path.append('.')
 def changelog_version(path, encoding='utf-8'):
     "Return the 1st *stable* (not pre, dev) version in the changelog"
     # See also grep_version() in setup.py
+    # e.g. "0.1.2, (1999-12-31)\n"
     version_pattern = re.compile(
-        r'^v(?P<version>[0-9]+\.[0-9]+\.[0-9]+)',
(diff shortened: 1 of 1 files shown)


moreati:
Revision 5ec61c90a664263f9849bf9d60b54c19a527ee67 also tried to use the flag `re.ASCII`, to restrict `\d` to ASCII digits, but this flag is unavailable in Python 2.x.

PR #1286, opened 2025-06-18
Keep compatibility with `setuptools` tagging wheels with `py2.py3`

### LINKED ISSUE #1283: setuptools bdist_wheel.universal is deprecated, removal expected 2025-08-30
Using `python -m build` to create a mitogen wheel results in several warnings, most critically that universal (i.e. Python 2 + Python 3) wheel support is due to end in a 2 months

```
!!

        ********************************************************************************
        With Python 2.7 end-of-life, support for building universal wheels
        (i.e., wheels that support both Python 2 and Python 3)
        is being obviated.
        Please discontinue using this option, or if you still need it,
        file an issue with pypa/setuptools describing your use case.

        By 2025-Aug-30, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.
        ********************************************************************************

!!
```

Mitogen currently relies on this.

[mitogen-build.log](https://github.com/user-attachments/files/20779504/mitogen-build.log)

TODO

- [ ] Is deprecation only within setuptools? Or will e.g. twine, pypi.org also reject universal wheels?

webknjaz:
Modern versions of `setuptools` emit a warning when the `universal = 1` option of `bdist_wheel` is used. This warning will turn into an error on Aug 30, 2025.

The only function of `universal = 1` is assigning the dual `py2.py3` tag to the wheels. It does not perform any content or metadata compatibility validation that might be related to this.

It is possible to keep producing same-tagged wheels by setting the non-deprecated `python_tag` option instead, which is what this PR does.

Fixes #1283

Ref https://github.com/pypa/setuptools/pull/4939

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [bdist_wheel]
-universal=1
+python_tag = py2.py3
 
 [coverage:run]
 branch = true


moreati:
Thank you. It'll probably be in a release later this month.

Issue #152, opened 2018-03-17
Setting ansible_python_interpreter within a play doesn't work


rouge8:
Here's an awful construct from one of my playbooks. The server has a Python app, with a virtualenv at `{{ venv_bin }}` that already has psycopg2 installed. Rather than also install psycopg2 systemwide, I set `ansible_python_interpreter` to point to that virtualenv for any tasks that need it:

```yaml
- name: Store the original ansible_python_interpreter
  set_fact: original_ansible_python_interpreter={{ ansible_python_interpreter }}

- name: Use app virtualenv for Postgres tasks
  set_fact: ansible_python_interpreter={{ venv_bin }}/python

- name: Ensure the app DB user exists
  postgresql_user:
    db: postgres
    login_host: "{{ db_credentials.host }}"
    port: "{{ db_credentials.port }}"
    login_user: "{{ db_superuser }}"
    login_password: "{{ db_superuser_password }}"
    name: "{{ db_credentials.user }}"
    password: "{{ db_credentials.password }}"
    state: present

- name: Return to default Ansible python
  set_fact: ansible_python_interpreter={{ original_ansible_python_interpreter }}
```

The only trick necessary for this to be supported by Ansible without mitogen is adding `ansible_python_interpreter: /usr/bin/python` to group/hostvars.

With mitogen, this errors: `FAILED! => {"changed": false, "msg": "the python psycopg2 module is required"}` which looks like it did not switch to the new `ansible_python_interpreter` variable. I'm okay with the answer being "this is terrible and unsupported" :).

Here's the mitogen logs from that task. mitogen does not log anything for the `set_fact` tasks:

```
[pid 27834] 12:02:40 D mitogen: Context(10, u'ssh.env1-stage-web').call_async(<function run_module at 0x110db4aa0>, *('ansible.modules.database.postgresql.postgresql_user',), **{ARGS OMITTED})
[pid 27818] 12:02:40 D mitogen: ModuleResponder(Router(Broker()))._on_get_module('ansible_mitogen')
[pid 27818] 12:02:40 D mitogen: _send_load_module(mitogen.ssh.Stream(u'ssh.env1-stage-web'), 'ansible_mitogen')
[pid 27818] 12:02:41 D mitogen: ModuleResponder(Router(Broker()))._on_get_module('ansible_mitogen.helpers')
[...]

PR #1299, opened 2025-07-10
parent: avoid early-EOF in sudo bootstrap, wait until full preamble is read

Forgetyk:
### Problem

When Mitogen is used with **`become: sudo`**, the first-stage bootstrap
sometimes receives **EOF on *stdin* before the compressed preamble arrives**.
Under Python 3.10 + Ansible 2.12 the sequence

```
MITO000
Traceback … TypeError: a bytes-like object is required, not 'NoneType'
```

is reproducible on every run.

* `_first_stage()` calls `fp.read(PREAMBLE_COMPRESSED_LEN)`
  → returns **`None`** instead of `bytes`.
* `zlib.decompress(None)` raises `TypeError`; controller reports
  “EOF on stream”.

Root cause: after writing the preamble the parent closes the write end of the
pipe immediately; with SSH compression (`ssh -C`) the tail of the deflate
stream can be lost, so the child sees premature EOF.

---

### Fix

Read the preamble **reliably**:

```python
remaining = PREAMBLE_COMPRESSED_LEN
chunks = []
deadline = time.time() + 10
while remaining:
    chunk = fp.read(remaining)
    if chunk:
        chunks.append(chunk)
        remaining -= len(chunk)
        continue
    if time.time() > deadline:
        raise RuntimeError('early EOF while reading preamble')
    time.sleep(0.02)
C = zlib.decompress(b''.join(chunks))
```

* loops until *all* bytes are received (or timeout hits);
* eliminates the `None` case, therefore no `TypeError`;
* zero functional impact on normal fast paths.

---

### DIFF
--- a/docs/ansible_detailed.rst
+++ b/docs/ansible_detailed.rst
@@ -1018,6 +1018,8 @@ Like the :ans:conn:`ssh` except connection delegation is supported.
 * ``ansible_ssh_private_key_file``
 * ``ansible_ssh_pass``, ``ansible_ssh_password``, ``ansible_password``
   (default: assume passwordless)
+* ``ansible_ssh_host_key_checking``, ``ansible_host_key_checking`` (default: 
+  :data:`True`)
 * ``ssh_args``, ``ssh_common_args``, ``ssh_extra_args``
 * ``mitogen_mask_remote_name``: if :data:`True`, mask the identity of the
   Ansible controller process on remote machines. To simplify diagnostics,
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -32,6 +32,10 @@ Running Ansible 12 + Mitogen will currently print a deprecation warning
 Ansible + Mitogen will still work for now. Mitogen is considering alternatives
 to strategy plugins under :gh:issue:`1278`.
 
+
+v0.3.25a3 (2025-07-02)
+----------------------
+
 * :gh:issue:`1285` CI: use `result_format = yaml` for Ansible test output,
   instead of deprecated `stdout_callback = yaml`
 * :gh:issue:`1293` CI: Fix ``ansible_version`` comparisons when an Ansible
--- a/mitogen/parent.py
+++ b/mitogen/parent.py
@@ -53,7 +53,7 @@ import termios
 import textwrap
 import threading
 import zlib
-
+import time
 # Absolute imports for <2.5.
 select = __import__('select')
 
@@ -1438,7 +1438,19 @@ class Connection(object):
             os.execl(sys.executable,sys.executable+'(mitogen:CONTEXT_NAME)')
         os.write(1,'MITO000\n'.encode())
         fp=os.fdopen(0,'rb')
-        C=zlib.decompress(fp.read(PREAMBLE_COMPRESSED_LEN))
+        remaining = PREAMBLE_COMPRESSED_LEN
+        chunks = []
+        deadline = time.time() + 10
+        while remaining:
+            chunk = fp.read(remaining)
+            if chunk:
+                chunks.append(chunk)
+                remaining -= len(chunk)
+                continue
+            if time.time() > deadline:
+                raise RuntimeError('early EOF while reading preamble')
+            time.sleep(0.02)
+        C = zlib.decompress(b''.join(chunks))
         fp.close()
         fp=os.fdopen(W,'wb',0)
         fp.write(C)
--- a/mitogen/parent.py
+++ b/mitogen/parent.py
@@ -1438,27 +1438,28 @@ class Connection(object):
             os.execl(sys.executable,sys.executable+'(mitogen:CONTEXT_NAME)')
         os.write(1,'MITO000\n'.encode())
         fp=os.fdopen(0,'rb')
+        import time
         remaining = PREAMBLE_COMPRESSED_LEN
         chunks = []
-        deadline = time.time() + 10
+        deadline = time.time() + 10.0
         while remaining:
             chunk = fp.read(remaining)
-            if chunk:
-                chunks.append(chunk)
-                remaining -= len(chunk)
+            if not chunk:
+                if time.time() >= deadline:
+                    break
+                time.sleep(0.02)
                 continue
-            if time.time() > deadline:
-                raise RuntimeError('early EOF while reading preamble')
-            time.sleep(0.02)
-        C = zlib.decompress(b''.join(chunks))
+            chunks.append(chunk)
+            remaining -= len(chunk)
+        try:
+            empty_bytes = bytes()
+        except NameError:
+            empty_bytes = ''
+        C = zlib.decompress(empty_bytes.join(chunks))
         fp.close()
-        fp=os.fdopen(W,'wb',0)
-        fp.write(C)
-        fp.close()
-        fp=os.fdopen(w,'wb',0)
-        fp.write(C)
-        fp.close()
-        os.write(1,'MITO001\n'.encode())
+        fp = os.fdopen(W, 'wb', 0); fp.write(C); fp.close()
+        fp = os.fdopen(w, 'wb', 0); fp.write(C); fp.close()
+        os.write(1, 'MITO001\n'.encode())
         os.close(2)
 
     def get_python_argv(self):


moreati:
There may be a bug here. A minmal reproducer would be needed to confim it. The proposed fix is incorrect, as shown by the failing test run.

Problems with the patch (non-exhaustive)
- Using `import time` at module level doesn't make `time` available in the first stage.
- `b''` is not available in Python < 2.6

Potential problems
- I'm unsure if `mitogen.parent.Connection.get_boot_command()` can handle multiple levels of indentation.

Notes to self
- `os.fdopen(0, 'wb', 0)` returns an `_io.FileIO` object (on Python 3.9)
- `io.FileIO` is unbuffered, inherits methods from `io.RawIOBase`
- [`io.RawIO.read(n)`](https://docs.python.org/3/library/io.html#io.RawIOBase.read)
  - Performs exactly one system call
  - Does not guarantee n bytes will be read
  - Returns `None` if `fp` is non-block, and no bytes are available.

moreati:
Scratch those notes to self.
- `_first_stage()` uses `os.fdopen(0,'rb')`, *not* `os.fdopen(0,'rb',0)`
- `os.fdopen(0,'rb')` returns an `_io.BufferedReader`
- [`io.BufferedIOBase.read(n)`](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read)
  - May perform multiple system calls
  - Also does not guarantee n bytes

Forgetyk:
I’ve pushed the latest amendments so that the CI suite now passes.
The remaining failures we saw (e.g. container-image pulls and legacy apt repository errors) were unrelated to the _first_stage() logic itself; they were caused by external factors (private GHCR images and Debian 10 archive URLs).

Please let me know if you’d like the fix split into smaller commits or if anything needs further clarification.

Thanks for the review!

moreati:
> There may be a bug here. A minmal reproducer would be needed to confim it.

I still haven't seen direct evidence that there is a bug. I'm making that a requirement - a reproducer that can form the basis of a regression test - before we move onto the design/finer points of any fix.

I'm looking at the Debian problem in CI.

Forgetyk:
Attached playbook startup log with error [ansible_log.txt](https://github.com/user-attachments/files/21221962/ansible_log.txt)

moreati:
> Attached playbook startup log with error [ansible_log.txt](https://github.com/user-attachments/files/21221962/ansible_log.txt)

As I said, I need a *reproducer*, a minimal playbook or Python script that exhibits the error. Something that anyone can run, and be adapted to become part of the test suite.

A log sample alone doesn't fufill that requirement.

moreati:
Notes to self
- `io.BufferedReader.read()` docs were recently updated, adding the note
  > When the underlying raw stream is non-blocking, implementations may either raise `BlockingIOError` or return `None` if no data is available. `io` implementations return None.
- The note was added in https://github.com/python/cpython/pull/130653, closing https://github.com/python/cpython/issues/80050

Open questions
1. Does `fp.read(PREAMBLE_COMPRESSED_LEN)` returning `None` indicate stdin has become non-blocking?
2. If so, how did that occur?
3. Are there other mechanisms that can cause it return `None`?

moreati:
@Forgetyk to help with ruling out a few possibilities
1. What version of  Python is your target (iac-test-2) running?
2. What is the output of `uname -a` on your target?

Alternatively, what is the output of ruinning a playbook like

```yaml
- hosts: iac-test-2
  become: true
  gather_facts: true
  strategy: linear
  tasks:
    - {debug: {var: ansible_facts.architecture}}
    - {debug: {var: ansible_facts.distribution}}
    - {debug: {var: ansible_facts.distribution_major_version}}
    - {debug: {var: ansible_facts.distribution_version}}
    - {debug: {var: ansible_facts.os_family}}
    - {debug: {var: ansible_facts.python}}
```

Forgetyk:
It is more difficult there the occurrence of this problem. In production it appears on all servers, but not on new virtual machines. Now I'm trying to determine exactly what is causing the problem, maybe it's a combination of the ssh settings of the version itself and the use of ssh parameters.

As soon as I get the exact result, I will report back.

moreati:
> @Forgetyk to help with ruling out a few possibilities
>     1. What version of  Python is your target (iac-test-2) running?
>     2. What is the output of `uname -a` on your target?

Please could you answer these for a host that has shown the failure. Even if it's not still doing so.

Forgetyk:
Was able to reproduce the problem and find the original source

---

## Environment

| Component          | Version  |
| ------------------ | ------------------------------------------ |
| OS | Linux 5247089-it57222 5.15.0-142-generic #152-Ubuntu SMP Mon May 19 10:54:31 UTC 2025 x86_64 x86_64 x86_64 x86_64 GNU/Linux |
| Python             |  3.10.12  |
| sudo               |  1.9.9 |
| OpenSSH            |  OpenSSH_8.9p1 Ubuntu-3ubuntu0.10, |
| ansible-core       |  2.12.10  |

All tests were run with **`strategy = mitogen_linear`** and `pipelining = true`.

---

## Reproducer

1. **Target host** – enable sudo I/O logging:

   ```/etc/sudoers
   Defaults log_output
   Defaults!/usr/bin/sudoreplay !log_output
   Defaults!/usr/local/bin/sudoreplay !log_output
   Defaults!/sbin/reboot !log_output
   Defaults loglinelen=0
   Defaults logfile=/var/log/sudo.log
   #Defaults:ansible !log_output, !log_input
   root    ALL=(ALL:ALL) ALL
   %admin ALL=(ALL) ALL
   %sudo   ALL=(ALL:ALL) ALL
   @includedir /etc/sudoers.d
   ```

2. **Target host** – ssh settings:

   ```/etc/ssh/sshd_config
	Protocol 2
	SyslogFacility AUTH
	LogLevel INFO
	LoginGraceTime 60
	PermitRootLogin no
	PermitEmptyPasswords no
	IgnoreRhosts yes
	HostbasedAuthentication no
	X11Forwarding no
	ClientAliveInterval 900
	ClientAliveCountMax 0
	AllowTcpForwarding no
	ChallengeResponseAuthentication no
	UsePAM yes
	Subsystem    sftp    /usr/lib/openssh/sftp-server
	PrintMotd no
	ListenAddress 0.0.0.
   ```

3. **Controller** – standard Mitogen config (only the interesting bits):

   ```ini
   # ansible.cfg
   [defaults]
   strategy_plugins = /…/ansible_mitogen/plugins/strategy
   strategy         = mitogen_linear
   [ssh_connection]
   pipelining = true
   ```

4. **Inventory**

   ```ini
   [target]
   iac-test-2 ansible_user=ansible ansible_python_interpreter=/usr/bin/python3
   ```

5. **Playbook**

   ```yaml
	  become: yes
	  gather_facts: no
	  hosts:
		- iac-test-2
	  roles:
		- some_role
	  vars:
		some_var: "{{ var }}"
   ```

6. **Result**

   ```
   BlockingIOError: [Errno 35] Resource temporarily unavailable
   mitogen.core.py", line 2247, in write
   ```

   The crash appears within a few hundred ms, long before any module code is executed.

---

### What is happening

* `sudo log_output/log_input` spawns a **pseudo‑terminal** in front of the command it runs and copies all I/O through that PTY.
  The PTY’s kernel buffer is only \~64 KiB.
* Mitogen opens both ends of its pipes **O\_NONBLOCK** and uses an event loop.
* During the bootstrap it pushes \~35 KiB of compressed preamble in one go; a couple of other writes follow immediately.
  When the PTY buffer fills up, `write()` returns **`EAGAIN` → `BlockingIOError`**, which is not retried and kills the stream.

Plain Ansible (no Mitogen) performs the same copy through sudo, **but its `os.write()` is blocking**, so it simply waits and survives.

---

## Work‑around that mask the issue

`Defaults:ansible !log_output, !log_input` might be a solution for other similar problems, but it is forbidden within our production regulations.

Because disabling I/O logging is off‑limits for compliance, we need Mitogen to **gracefully retry `EAGAIN` (or temporarily switch to blocking writes)**.

moreati:
Thank you thats great dtective/dignostic work. I've raised #1306 based on it. I'm not sure if the ultimate should be in `_first_stage()` elsewhere, or both. I'll try to recreate the error locally, so we can have a test case.

moreati:
> | Component | Version |
> | -- | -- |
> | OS | Linux 5247089-it57222 5.15.0-142-generic # 152-Ubuntu SMP Mon May 19 10:54:31 UTC 2025 x86_64 x86_64 x86_64 x86_64 GNU/Linux |
> | Python | 3.10.12 |
> | sudo | 1.9.9 |
> | OpenSSH | OpenSSH_8.9p1 Ubuntu-3ubuntu0.10 |
> | ansible-core | 2.12.10 |

@Forgetyk could you confirm
1. Is this the same on the Ansible controller *and* the targets? Or do the differ? In particular do the Python versions differ?

Edit: following is notes to myself:

This is Ubuntu 22.04 (Jammy), going by the kernel, sudo and OpenSSH version. The Python version also tallies with recent updates. At time of writing Python 3.10.18 is the latest 3.10 release https://docs.python.org/release/3.10.18/whatsnew/changelog.html

Forgetyk:
All specified versions are the same on the controller and target hosts

moreati:
I'm still trying to reproduce this under #1306

1. What version of Mitogen are you using?
2. If it is < 0.3.24 please could you try Mitogen 0.3.24 or 0.3.25b1?

Forgetyk:
I have used versions 0.3.24 and 0.3.25a3. on both versions the problem is reproducible. I can test 0.3.25b1 if needed.

moreati:
> I can test 0.3.25b1 if needed.

No need. There was a moderate chance 0.3.24 contained releavant changes, a slim chance one of the alphas did. There's no chance the change(s) in b1 is/are relevant. Thanks for confirming.

moreati:
@Forgetyk I've create #1307 with an alternative fix, and regression tests. I've you get a chance please could you try it. Thanks, and sorry for the wait. Alex

moreati:
Superceded by #1307

Issue #712, opened 2020-04-22
Sporadic EAGAIN errors due to non-blocking stdout


apyrgio:
When using Mitogen, I sometimes see `EAGAIN` errors, due to stdout being non-blocking. This breaks the Python interpreter, as well as any other spawned commands, since they don't handle non-blocking file descriptors. To trigger this, one can simply do `sys.stdout.write("A" * 1000000)` on the remote, which will fill the kernel buffer and return `EAGAIN`, instead of blocking.

Reading the sections on ["Standard IO Redirection"] and ["The IO Multiplexer"], my understanding is that Mitogen replaces stdout with a Unix pipe. It gives the write end of the Unix pipe to the Python interpreter, and the read end to a poller (the I/O multiplexer). I understand that the read end typically needs to be non-blocking, else the I/O multiplexer is not fully async. However, I don't see why the write end (what the interpreter perceives as stdout) needs to be non-blocking as well, since it breaks the assumptions of many programs.

So, my question is, is stdout non-blocking by design, or is it a bug? If it's the former, what's the rationale behind it?

["Standard IO Redirection"]: https://mitogen.networkgenomics.com/howitworks.html#standard-io-redirection
["The IO Multiplexer"]: https://mitogen.networkgenomics.com/howitworks.html#the-io-multiplexer

dw:
This is/was an old bug at least in the Ansible extension, I definitely remember fixing it and even have an inkling of writing a test for it.

If you're using the core library, it's possible the problem was fixed in the wrong place, which is why you're still seeing it

apyrgio:
Hi @dw. Quick question, did you manage to reproduce it with the above script?

moreati:
Confirm this is still a problem in latest Mitogen. #1185 may or may not be related. 

```python
import argparse
import sys

import mitogen


def shout(num):
    sys.stdout.write("A" * num)


@mitogen.main()
def main(router):
    parser = argparse.ArgumentParser()
    parser.add_argument('-p', '--python', default='python3')
    parser.add_argument('hostname')
    parser.add_argument('num', type=int)
    args = parser.parse_args()

    context = router.ssh(hostname=args.hostname, python_path=args.python)
    context.call(shout, args.num)

    print("Passed. Try with a larger payload, e.g., 1000000")
```

```console
➜  mitogen git:(issue1185-blockingioerror) ✗ python3 issue713_repro.py pve.lan 1_000_000
[...]

PR #1300, opened 2025-07-10
stdio EAGAIN investigation

moreati:
refs #712

See
- https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_258
- https://biriukov.dev/docs/fd-pipe-session-terminal/1-file-descriptor-and-open-file-description/


moreati:
TODO

- [x] API: Should `Side.accept()` have a `blocked` argument, or something else?
- [x] Does it makes sense to leave the receive side (`rsock`) in non-blocking mode?
- [x] Better test string than `'A' * 2**20`?

PR #1341, opened 2025-10-23
master: Fix LogForwarder in case an own LogRecordFactory is used

mhartmay:
Since Python 3.2 the log record factory can be changed by using `logging.setLogRecordFactory` [1]. Therefore use `logging.makeLogRecord` as recommended in the documentation:

"LogRecord instances are created automatically by the Logger every time something is logged, and can be created manually via makeLogRecord() (for example, from a pickled event received over the wire)." [2]

This fixes the following error for my use case:

--- Logging error ---
  File "/usr/lib/python3.12/logging/__init__.py", line 464, in format
    return self._format(record)
           ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/logging/__init__.py", line 460, in _format
    return self._fmt % values
           ~~~~~~~~~~^~~~~~~~
KeyError: 'relpathname'

[1] https://docs.python.org/3/library/logging.html#logging.setLogRecordFactory
[2] https://docs.python.org/3/library/logging.html#logrecord-objects

### DIFF
--- a/mitogen/master.py
+++ b/mitogen/master.py
@@ -370,15 +370,19 @@ class LogForwarder(object):
         if logger is None:
             self._cache[logger_name] = logger = logging.getLogger(logger_name)
 
+        levelno = int(level_s)
         # See logging.Handler.makeRecord()
-        record = logging.LogRecord(
-            name=logger.name,
-            level=int(level_s),
-            pathname='(unknown file)',
-            lineno=0,
-            msg=s,
-            args=(),
-            exc_info=None,
+        record = logging.makeLogRecord(
+            {
+                "name": logger.name,
+                "levelname": logging.getLevelName(levelno),
+                "levelno": levelno,
+                "pathname": "(unknown file)",
+                "lineno": 0,
+                "msg": s,
+                "args": (),
+                "exc_info": None,
+            }
         )
         record.mitogen_message = s
         record.mitogen_context = self._router.context_by_id(msg.src_id)


mitogen/master.py
```diff
@@ -370,15 +370,19 @@ def _on_forward_log(self, msg):
...
-            level=int(level_s),
-            pathname='(unknown file)',
-            lineno=0,
-            msg=s,
-            args=(),
-            exc_info=None,
+        record = logging.makeLogRecord(
+            {
+                "name": logger.name,
+                "levelname": logging.getLevelName(levelno),
```
moreati: Nitpick: the old code didn't explicitly reinterpret the levelno to a level name, but there's a high chance it happened implicitly.

moreati: It did https://github.com/python/cpython/blob/4556e7e002fb21ddd9e4b16c7266877a4d99ea42/Lib/logging/__init__.py#L229

mitogen/master.py
```diff
@@ -370,15 +370,19 @@ def _on_forward_log(self, msg):
...
         # See logging.Handler.makeRecord()
-        record = logging.LogRecord(
-            name=logger.name,
-            level=int(level_s),
-            pathname='(unknown file)',
-            lineno=0,
-            msg=s,
-            args=(),
-            exc_info=None,
+        record = logging.makeLogRecord(
```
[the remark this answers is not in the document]
mhartmay: Is that okay? Or do we need to support even older versions?

mhartmay:
Thanks for merging!

I missed to add a changelog entry.... shall I add it or will you do?

moreati:
Thanks for the fix. I did the changelog in https://github.com/mitogen-hq/mitogen/pull/1346

PR #1307, opened 2025-07-21
mitogen: Fix non-blocking IO errors in first stage of bootstrap

moreati:
When /etc/sudoers has log_output (or similar) enabled the process spawned by `ctx.sudo()` via `mitogen.parent.Connection.start_child()` receives a stdin that is in non-blocking mode. The immediate symptom is that `os.openfd(0, ...).read(n)` sometimes returns `None`, causing the first stage to raise an unhandled `TypeError`.

The fix (for now) is to use `select.select()` in as while loop to read stdin. This increases the command size slightly, but I think it's a reasonable tradeoff until/unless the cause is more fully understood.

All CI tests are now run with sudoers log_output enabled, in order to catch regressions. `first_stage_test.CommandLineTest` has been amended, because it relied on implementation details of the bootstrap process that are no longer true.

Fixes #1306, thanks to @Forgetyk for their detailed error reporting, diagnosis, and initial PR #1299

### Before
```
SSH command size: 755
Preamble (mitogen.core + econtext) size: 18227 (17.80KiB)

                        Original           Minimized           Compressed
mitogen.core         152218 148.7KiB  68437 66.8KiB 45.0%  18124 17.7KiB 11.9%
mitogen.parent        98853  96.5KiB  51103 49.9KiB 51.7%  12881 12.6KiB 13.0%
mitogen.fork           8445   8.2KiB   4139  4.0KiB 49.0%   1652  1.6KiB 19.6%
mitogen.ssh           10827  10.6KiB   6893  6.7KiB 63.7%   2099  2.0KiB 19.4%
mitogen.sudo          12089  11.8KiB   5924  5.8KiB 49.0%   2249  2.2KiB 18.6%
mitogen.select        12325  12.0KiB   2929  2.9KiB 23.8%    964  0.9KiB  7.8%
mitogen.service       41581  40.6KiB  22398 21.9KiB 53.9%   5847  5.7KiB 14.1%
mitogen.fakessh       15767  15.4KiB   8149  8.0KiB 51.7%   2676  2.6KiB 17.0%
mitogen.master        55317  54.0KiB  28846 28.2KiB 52.1%   7528  7.4KiB 13.6%
```

### After
```
SSH command size: 798
Preamble (mitogen.core + econtext) size: 18227 (17.80KiB)

                        Original           Minimized           Compressed     
mitogen.core         152218 148.7KiB  68437 66.8KiB 45.0%  18124 17.7KiB 11.9%
[...]

Issue #1306, opened 2025-07-21
Mitogen + sudo + log_stdin/log_stdout leads to preamble read error in first stage (EAGAIN, EWOULDBLOCK, None)


moreati:
Under certain conditions (e.g. `sudo` with the `log_input` or `log_output` flag) the stdin of Mitogen's first stage can be set non-blocking (`O_NONBLOCK`) this results in `fp.read()` returning `None` if the buffer is empty.

The error seen by the user looks like

```
MITO000
Traceback … TypeError: a bytes-like object is required, not 'NoneType'
```

This may be related to #712, #1185.

Credit to @forgetyk for the original error and diagnosis, their words follow.

> ## Environment
> 
> | Component          | Version  |
> | ------------------ | ------------------------------------------ |
> | OS | Linux 5247089-it57222 5.15.0-142-generic #152-Ubuntu SMP Mon May 19 10:54:31 UTC 2025 x86_64 x86_64 x86_64 x86_64 GNU/Linux |
> | Python             |  3.10.12  |
> | sudo               |  1.9.9 |
> | OpenSSH            |  OpenSSH_8.9p1 Ubuntu-3ubuntu0.10, |
> | ansible-core       |  2.12.10  |
> 
> All tests were run with **`strategy = mitogen_linear`** and `pipelining = true`.
> 
> ## Reproducer
> 
> 1. **Target host** – enable sudo I/O logging:
> 
>    ```
>    # /etc/sudoers
>    Defaults log_output
>    Defaults!/usr/bin/sudoreplay !log_output
>    Defaults!/usr/local/bin/sudoreplay !log_output
>    Defaults!/sbin/reboot !log_output
>    Defaults loglinelen=0
>    Defaults logfile=/var/log/sudo.log
>    #Defaults:ansible !log_output, !log_input
>    root    ALL=(ALL:ALL) ALL
>    %admin ALL=(ALL) ALL
>    %sudo   ALL=(ALL:ALL) ALL
>    @includedir /etc/sudoers.d
>    ```
> 
> 2. **Target host** – ssh settings:
> 
>    ```
>    # /etc/ssh/sshd_config
> 	Protocol 2
> 	SyslogFacility AUTH
> 	LogLevel INFO
> 	LoginGraceTime 60
> 	PermitRootLogin no
> 	PermitEmptyPasswords no
> 	IgnoreRhosts yes
> 	HostbasedAuthentication no
> 	X11Forwarding no
[...]

PR #1349, opened 2025-10-30
mitogen: Prevent hung bootstrap processes, add 5   second timeout to first stage

moreati:
Since using select.select() in the first stage (to handle an obscure corner case where stdin appears to be non-blocking) there has been a report of first stage processes running for ever in an infinite loop - reading 0 bytes from stdin.

This attempts to do an end run around that problem by aborting if the bootstrap takes longer than a few seconds for *any* reason. Existing retry logic should deal with it as before.

refs #1306, #1307, #1348

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -21,6 +21,9 @@ To avail of fixes in an unreleased version, please download a ZIP file
 In progress (unreleased)
 ------------------------
 
+* :gh:issue:`1266` :mod:`mitogen`: Prevent hung bootstrap processes, add 5
+  second timeout to first stage
+
 
 v0.3.34 (2025-11-27)
 --------------------
--- a/mitogen/parent.py
+++ b/mitogen/parent.py
@@ -1436,6 +1436,10 @@ class Connection(object):
             if os.uname()[0]+os.uname()[2][:2]+sys.version[:3]=='Darwin212.7':os.environ['PYTHON_LAUNCHED_FROM_WRAPPER']='1'
             os.environ['ARGV0']=sys.executable
             os.execl(sys.executable,sys.executable+'(mitogen:%s)'%sys.argv[2])
+        # Timeout execution after a few seconds, to prevent hung processes.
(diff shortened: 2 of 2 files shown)


moreati:
An experiment to test my belief about how this patch will behave. The following code is `fork_it.py` a simplified version of the first stage

```python
#!/usr/bin/env python3

import os
import signal
import sys

R, W = os.pipe()

if child_pid := os.fork():
    os.dup2(0, 100)
    os.dup2(R, 0)
    os.close(R)
    os.close(W)
    sys.stdout.write(f'parent: {os.getpid()}, child: {child_pid}\n')
    os.execl(sys.executable, sys.executable)
else:
    signal.alarm(2)
    s = sys.stdin.read() 
    with os.fdopen(W, 'wb', 0) as f:
        f.write(b'print("Hello world!")\n')
```

The simplifications include
- The fork child sends a hard coded preamble (`b'print("Hello world!")\n'`)
- The fork parent discards whatever it reads from stdin, this is intentional.
- stdin can be safely assumed to be blocking
- There is no compression, or stage 0.
- There is only a single pipe, since there's no need to cache a copy of the preamble
- For convenience the SIGALM timeout is only 2 seconds.

What I expect:
- When no data is provided to stdin
  - the call to sys.stdin.read() blocks
  - eventually the fork child is killed by SIGALRM
  - when the child terminates the python interpreter in the parent will receive EOF and also terminate itself
  - the simulated preamble will never be received, so won't execute or print anything
  - no lingering processess should be left

- When data is provided
  - `Hello world!` will be printed
  - both parent and child will exit gracefully

Results appear to match this

```console
➜  mitogen git:(master) ✗ python3 fork_it.py        
parent: 11482, child: 11483
➜  mitogen git:(master) ✗ ps -p 11482,11483 
  PID TTY           TIME CMD
```
```console
➜  mitogen git:(master) ✗ python3 fork_it.py <<< foo
parent: 11725, child: 11726
Hello world!
```

Issue #1360, opened 2025-11-13
atexit wrapper orphans temp files with kubernetes.core.k8s ansible module


dmurnane:
Environment:
Rocky Linux 9.6
python3-3.9.21-2.el9_6.2.x86_64
python3-mitogen-0.3.22-1.noarch
ansible-core-2.14.18-1.el9.x86_64
python3-kubernetes-26.1.0-3.el9.noarch

The python kubernetes module creates temp files to hold certificates to communicate with the kubernetes API server. It registers an atexit handler for `kubernetes.config.kube_config._cleanup_temp_files` in order to remove those files on exit. It appears that the mitogen AtExitWrapper is masking that cleanup, so these files accumulate over time when kubernetes.core.k8s tasks are run. `mitogen_task_isolation: fork` does not resolve the issue, but switching to `strategy: linear` does properly clean up the files.

Patching _atexit__register in ansible_mitogen/runner.py with `if func == shutil.rmtree or func == kubernetes.config.kube_config._cleanup_temp_files:` appears to clean up the files but this feels rather more fragile than is reasonable.

Reproduction:
```
[dmurnane@localhost ~]$ ls /tmp/tmp*
ls: cannot access '/tmp/tmp*': No such file or directory
[dmurnane@localhost ~]$ cat k8s-test.yml
# code: language=ansible
---
- name: Testing
  hosts: localhost
  tasks:
    - name: Get coredns deployment
      kubernetes.core.k8s_info:
        api_version: apps/v1
        kind: Deployment
        name: coredns
        namespace: kube-system
[dmurnane@localhost ~]$ ansible-playbook k8s-test.yml

PLAY [Testing] ****************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Get coredns deployment] *************************************************************************************************************************************************************************************
ok: [localhost]

PLAY RECAP ********************************************************************************************************************************************************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0

[dmurnane@localhost ~]$ ls /tmp/tmp*
/tmp/tmpte9emays  /tmp/tmputdhak8k  /tmp/tmpwtgoybri
```

moreati:
@dmurnane 
1. Was there a time that these files got cleaned up? I.e. has this only started occuring recently?
2. What version of the Kubernetes collection are you running?

Questions/notes to self
[...]

PR #1388, opened 2025-12-04
WIP: atexit handling in Ansible modules

moreati:
Pin down if (and when) callbacks added by `atexit.register()` are executed. Fix or document shortcomings.

refs #1360

### DIFF
--- a/tests/ansible/lib/modules/atexit_cleanup.py
+++ b/tests/ansible/lib/modules/atexit_cleanup.py
@@ -1,0 +1,43 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+import atexit
+import errno
+import os
+
+from ansible.module_utils.basic import AnsibleModule
+
+CLEANUP_FILE = '/tmp/mitogen_test_atexit_cleanup_canary.txt'
+
+
+def cleanup(file):
+    try:
+        os.unlink(file)
+    except OSError as exc:
+        if exc.errno == errno.ENOENT:
+            pass
+        raise
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec={},
+    )
+
+    atexit.register(cleanup, CLEANUP_FILE)
+
+    with open(CLEANUP_FILE, 'wb') as f:
+        f.truncate()
+
+    result = {
+        'changed': True,
+    }
+    module.exit_json(**result)
+
+
+
+if __name__ == '__main__':
+    main()
--- a/tests/ansible/regression/all.yml
+++ b/tests/ansible/regression/all.yml
@@ -19,3 +19,4 @@
 - import_playbook: issue_1066__add_host__host_key_checking.yml
 - import_playbook: issue_1079__wait_for_connection_timeout.yml
 - import_playbook: issue_1087__template_streamerror.yml
+- import_playbook: issue_1360_atexit_cleanup.yml
--- a/tests/ansible/regression/issue_1360_atexit_cleanup.yml
+++ b/tests/ansible/regression/issue_1360_atexit_cleanup.yml
@@ -1,0 +1,34 @@
+- name: regression/issue_1360_atexit_cleanup.yml
+  hosts: test-targets
+  gather_facts: false
+  become: false
+  vars:
+    cleanup_canary_path: /tmp/mitogen_test_atexit_cleanup_canary.txt
+  tasks:
+    - block:
+        - name: Remove lingering cleanup canary
+          file:
+            path: "{{ cleanup_canary_path }}"
+            state: absent
+
+        - name: Run module with atexit cleanup
+          atexit_cleanup:
+
+        - name: Trigger exit of remote process
+          meta: reset_connection
+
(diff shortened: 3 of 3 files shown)


moreati:
I expect the test to fail, based on
```console
➜  ansible git:(issue1360) ✗ ANSIBLE_STRATEGY=linear ansible localhost -m atexit_cleanup 
[WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details
localhost | CHANGED => 
    changed: true
➜  ansible git:(issue1360) ✗ ls -l /tmp/mitogen_test*
zsh: no matches found: /tmp/mitogen_test*
```
```console
➜  ansible git:(issue1360) ✗ ANSIBLE_STRATEGY=mitogen_linear ansible localhost -m atexit_cleanup
[WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details
localhost | CHANGED => 
    changed: true
➜  ansible git:(issue1360) ✗ ls -l /tmp/mitogen_test*                                           
-rw-r--r--@ 1 alex  wheel  0 Dec  4 09:42 /tmp/mitogen_test_atexit_cleanup_canary.txt
```

moreati:
In old CPython (probably 2.x) atexit was a pure Python module. The list of registered callbacks was stored in `atexit._exithandlers`.

In current CPython (since https://github.com/python/cpython/commit/670e6921349dd408b6958a0c5d3b1486725f9beb) atexit is a builtin module and the list of handlers is stored internally. I don't see a way to introspect it. The number of handlers can be introspected, and maybe > 0 depending how CPython was started

```
$ python3.14 -S
Python 3.14.0 (main, Oct 14 2025, 21:10:22) [Clang 20.1.4 ] on darwin
>>> import atexit; atexit._ncallbacks()
1

$ python3.14   
Python 3.14.0 (main, Oct 14 2025, 21:10:22) [Clang 20.1.4 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import atexit; atexit._ncallbacks()
2

$ python3.14 -c "import atexit; print(atexit._ncallbacks())"
0
```

moreati:
From `-vvv` output it looks like the process(es) for the target are being terminated by SIGTERM (request graceful process shutdown)

> [mux  53761] 11:42:47.504339 D mitogen.parent: Router(Broker(2510)): deleting route to 3
> [mux  53761] 11:42:47.504575 D mitogen.[local.53763]: MitogenProtocol(fork.53765): disconnecting
> [mux  53761] 11:42:47.540454 D mitogen.parent.[local.53763]: Process fork.53765 pid 53765: exited due to signal 15 (SIGTERM)
> [mux  53761] 11:42:47.540868 D mitogen.[local.53763]: Broker(01a0): force disconnecting <Side of parent fd 5>
> [mux  53761] 11:42:47.541144 D mitogen.[local.53763]: parent stream is gone, dying.
> [mux  53761] 11:42:47.541417 D mitogen.[local.53763]: Broker(01a0): shutting down
> [mux  53761] 11:42:47.541711 D mitogen: <Side of local.53763 fd 105>: empty read, disconnecting
> [mux  53761] 11:42:47.542019 D mitogen.parent: PopenProcess local.53763 pid 53763: exited due to signal 15 (SIGTERM)
> [mux  53761] 11:42:47.542289 I ansible_mitogen.services: ContextService(): Forgetting Context(2, 'local.53763') due to stream disconnect
> [mux  53761] 11:42:47.542549 D mitogen.route_monitor: stream local.53763 is gone; propagating DEL_ROUTE for {2}

`atexit` doesn't run handlers in that case (emphasis mine)
> functions registered via this module are not called when **the program is killed by a signal not handled by Python**, when a Python fatal internal error is detected, or when [os._exit()](https://docs.python.org/3.10/library/os.html#os._exit) is called.

moreati:
Two third party packages worth perusing
1. https://pypi.org/project/multiexit/
   > A better, saner and more useful [atexit](https://docs.python.org/3/library/atexit.html) replacement for Python 3 that supports [multiprocessing](https://docs.python.org/3/library/multiprocessing.html).
2. https://pypi.org/project/safe-exit/
   > Safe Exit is a Python package that provides functionality to handle graceful process termination. The package allows users to register functions that will be called when the program exits.

The limitation is that third-party packages (e.g. kubernetes client) are already using these, so they may not be applicable as is.

moreati:
Reproduced with just Mitogen
```python
import atexit
import sys
import os

import mitogen.master


def cleanup(file):
    try:
        os.unlink(file)
    except FileNotFoundError:
        pass


def foo(file):
    atexit.register(cleanup, file)
    with open(file, 'wb') as f:
        f.truncate()
    return 42


if __name__ == '__main__':
    broker = mitogen.master.Broker()
    router = mitogen.master.Router(broker)
    context = router.local(python_path=sys.executable)

    canary = '/tmp/issue1360_canary.txt'
    cleanup(canary)

    result = context.call(foo, canary)
    print(result)
```
```console
➜  mitogen git:(issue1360) ✗ python3 issue1360.py
42
➜  mitogen git:(issue1360) ✗ ls /tmp/issue1360_canary.txt                  
/tmp/issue1360_canary.txt
```

PR #1389, opened 2025-12-04
mitogen: first_stage: Break the while loop in case of EOF

mhartmay:
The current implementation can cause an infinite loop, leading to a process that hangs and consumes 100% CPU. This occurs because the EOF condition is not handled properly, resulting in repeated select(...) and read(...) calls.

The fix is to properly handle the EOF condition and break out of the loop when it occurs.

Fixes: https://github.com/mitogen-hq/mitogen/issues/1348

### DIFF
--- a/mitogen/parent.py
+++ b/mitogen/parent.py
@@ -1436,11 +1436,16 @@ class Connection(object):
             os.environ['ARGV0']=sys.executable
             os.execl(sys.executable,sys.executable+'(mitogen:%s)'%sys.argv[2])
         os.write(1,'MITO000\n'.encode())
+        # Size of the compressed core source to be read
+        n=int(sys.argv[3])
         # Read `len(compressed preamble)` bytes sent by our Mitogen parent.
         # `select()` handles non-blocking stdin (e.g. sudo + log_output).
         # `C` accumulates compressed bytes.
         C=''.encode()
-        while int(sys.argv[3])-len(C)and select.select([0],[],[]):C+=os.read(0,int(sys.argv[3])-len(C))
+        # data chunk
+        V='V'
+        # Stop looping if no more data is needed or EOF is detected (empty bytes).
+        while n-len(C) and V:select.select([0],[],[]);V=os.read(0,n-len(C));C+=V
         # Raises `zlib.error` if compressed preamble is truncated or invalid
         C=zlib.decompress(C)
         f=os.fdopen(W,'wb',0)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -27,6 +27,7 @@ In progress (unreleased)
 * :gh:issue:`1354` docs: Document Ansible 13 (ansible-core 2.20) support
 * :gh:issue:`1354` :mod:`mitogen`: Clarify error message when a module
   request would be refused by allow or deny listing
+* :gh:issue:`1348` :mod:`mitogen`: Fix hanging process with 100% CPU usage
 
 
 v0.3.35 (2025-12-01)


mhartmay:
**Note:** I did not test whether this actually fixes the problem since I did not reproduce the problem, but it's very likely as I had similar issues in the past.

mhartmay:
Question: Should we raise an error if not all data could be read?

Update: What speaks against raising an error is that it would be different to the "original code" behavior (before https://github.com/mitogen-hq/mitogen/pull/1307)

The original code reads:
```python
        fp=os.fdopen(0,'rb')
        C=zlib.decompress(fp.read(PREAMBLE_COMPRESSED_LEN))
        fp.close()
```

`fdopen` returns an `BufferedReader` and the documentation for `fdopen(..).read()` says:

> Fewer bytes may be returned than requested. [1]

So even with the old version it was possible that less than expected data was read and no exception was raised, so we should probably keep to that behavior for now.

[1] https://docs.python.org/3/library/io.html#io.BufferedIOBase.read

moreati:
> Question: Should we raise an error if not all data could be read?

The call to `zlib.decompress()` raises an exception if the data is truncated.

mitogen/parent.py
```diff
@@ -1437,8 +1439,8 @@ def _first_stage():
             os.environ['ARGV0']=sys.executable
             os.execl(sys.executable,sys.executable+'(mitogen:%s)'%sys.argv[2])
         os.write(1,'MITO000\n'.encode())
-        C=''.encode()
-        while int(sys.argv[3])-len(C)and select.select([0],[],[]):C+=os.read(0,int(sys.argv[3])-len(C))
+        n=int(sys.argv[3]);C=''.encode();V='V'
+        while n>len(C) and V:select.select([0],[],[]);V=os.read(0,n-len(C));C+=V
```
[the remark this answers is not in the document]
mhartmay: Wouldn't we get an BlockingIOError [1] in that case? At least the documentation of `os.read` [2] reads:

> Return a bytestring containing the bytes read. If the end of the file referred to by fd has been reached, an empty bytes object is returned.

[1] https://docs.python.org/3/library/exceptions.html#BlockingIOError
[2] https://docs.python.org/3/library/os.html#os.read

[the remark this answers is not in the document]
mhartmay: Yes, but we use `os.read(...)`  not the `BufferedIO` `read` (e.g. `open(...).read()`).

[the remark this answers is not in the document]
mhartmay: > Hmm, I need to need resh my memory of #1306 and #1307

`fp.read()` <--- that's the `bufferedIO read != os.read(..)`

and the original code used BufferedIO as `os.fdopen` is more or less an alias for `open(...)` [1] :)
`
  fp=os.fdopen(0,'rb')
        C=zlib.decompress(fp.read(PREAMBLE_COMPRESSED_LEN))
`

[1] https://docs.python.org/3/library/os.html#os.fdopen

mhartmay:
@moreati I've added timeout handling, but I'm not sure which kind of exception should be raised in case of a timeout (currently, I've `Exception('TODO')`. And I would rather use a large timeout as we do not know how fast the systems are.

Edit: Maybe even make the timeout value configurable.

mhartmay:
> @moreati I've added timeout handling, but I'm not sure which kind of exception should be raised in case of a timeout (currently, I've `Exception('TODO')`. And I would rather use a large timeout as we do not know how fast the systems are.
> 
> Edit: Maybe even make the timeout value configurable.

I've added some preliminary steps to make it configurable and added a test for the timeout situation.

moreati:
By all means continue experimenting if you wish to, but be aware I'm unlikely to accept many new features into this chunk of code - particularly if they increase the size. My preference is to keep the first stage as minimal is possible.

To measure the size (after encoding and compression) use https://github.com/mitogen-hq/mitogen/blob/master/preamble_size.py.

I'm currently doing my experiments to pin down how `{fobj,os}.read()` behave in different circumstancess and across supported Python versions.

mhartmay:
Yep, right now I'm experimenting, especially in regard with the tests. Because it would be great if we have minimal tests for blocking and non-blocking STDIO. So the timeout test should be useful in all cases even for your SIGALRM solution :)

mhartmay:
> By all means continue experimenting if you wish to, but be aware I'm unlikely to accept many new features into this chunk of code - particularly if they increase the size. My preference is to keep the first stage as minimal is possible.

Fine with me :) Reason why I'm thinking about making the timeout configurable is that the halting problem is undecidable and the timeout depends on the system/environment.

> 
> To measure the size (after encoding and compression) use https://github.com/mitogen-hq/mitogen/blob/master/preamble_size.py.

Will add the diff to all commits then :) Do we care more about the SSH command size? preamble? Or mitogen.parent? Or all of them?

> 
> I'm currently doing my experiments to pin down how `{fobj,os}.read()` behave in different circumstancess and across supported Python versions.

Good!

moreati:
> Will add the diff to all commits then :) Do we care more about the SSH command size? preamble? Or mitogen.parent? Or all of them?

One before and after for a complete PR will be fine. SSH command size is what I minimize.

mhartmay:
@moreati I've now added two test cases, one that causes an timeout and the other that causes the EOF situation similar to the original bug.

moreati:
Based on https://github.com/moreati/nonblock_lab/commit/c4938b9e927a87b0a91b6ffddc4197b6bdc77fcf
- other end of a fifo/pipe is closed (never opened)
  - Python 2.x and Pytrhon 3.x `<x>.read()` both return empty byte string
- other end is open, but empty
  - Python 2.x `<x>.read()` raises an EAGAIN/EWOULDBLOCK derived exception
  - Python 3.x `fobj.read()` returns `None`
  - Python 2.x `os.read()` raises `BlockingIOError`

mhartmay:
@moreati 
I've now added one test case that runs the full stage code (`tests.first_stage_test.CommandLineTest.test_stage`).

I'm going to add multiple other test cases, e.g. 
stdin is a blocking FD, stdin is non-blocking, different pipe sizes, and maybe even passing a FD reading from a file.

e.g.

```
with open("input.file", "rb") as f: 
    proc = subprocess.Popen(
        ...
        stdin=f,
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
    )
    out, err = proc.communicate()
```

Do you have other ideas? Not sure if the file case if even needed...

Edit: To me it seems like a good idea to add more test cases to `first_stage_test.py` as we have everything under control that we want to test and it's easier to debug than a larger stack.

mhartmay:
@moreati 
Is there a difference to run test `first_stage_test` via ./run_tests.sh or manually via Python unittest executor?

That's the command line I'm using during development:
```
$ PYTHONPATH=$(pwd)/tests:$PYTHONPATH python -m unittest -v tests.first_stage_test.CommandLineTest
test_eof_too_early (tests.first_stage_test.CommandLineTest.test_eof_too_early)
The boot command should write an ECO marker to stdout, read the ... ok
test_stage (tests.first_stage_test.CommandLineTest.test_stage)
Test that first stage works ... ok
test_stdin_blocking (tests.first_stage_test.CommandLineTest.test_stdin_blocking)
Test that first stage works with blocking STDIN ... ok
test_stdin_non_blocking (tests.first_stage_test.CommandLineTest.test_stdin_non_blocking)
Test that first stage works with non-blocking STDIN ... ok
test_timeout_error (tests.first_stage_test.CommandLineTest.test_timeout_error)
The boot command should write an ECO marker to stdout, read the ... ok
test_valid_syntax (tests.first_stage_test.CommandLineTest.test_valid_syntax)
Test valid syntax ... ok

Ran 6 tests in 0.789s

OK
```

mhartmay:
> @moreati Is there a difference to run test `first_stage_test` via ./run_tests.sh or manually via Python unittest executor?

Okay, I found one difference:

MITOGEN_LOG_LEVEL=debug is enabled, that results in a different output. Will change the test accordingly.

moreati:
> Is there a difference to run test `first_stage_test` via ./run_tests.sh or manually via Python unittest executor?

Almost certainly yes, but I don't remember exactly how. Probably something to do with what $CWD is set to, how Pythons sys.path is constructed and whther test modules are imported as foo_test or tests.foo_test. I have resigned myself to letting GH actions run the test suite, until I can overhaul it.

tests/first_stage_test.py
```diff
@@ -1,11 +1,15 @@
+import fcntl
+import os
+import select
 import subprocess
```
moreati: Try replacing this with`testlib.subprocess`, it provides https://pypi.org/project/subprocess32/ as necessary, which backports the timeout features added in Python 3.3.

tests/first_stage_test.py
```diff
@@ -16,37 +20,322 @@ class CommandLineTest(testlib.RouterMixin, testlib.TestCase):
...
+            try:
+                returncode = proc.wait(timeout=10)
+            except TypeError:
+                returncode = proc.wait()
+        except:
+            proc.kill()
+            self.fail("First stage did not finish")
+        else:
+            try:
+                # proc was killed by SIGTERM?!? TODO Where does this come from?
```
moreati: Probably similar or same cause as #1360

tests/first_stage_test.py
```diff
@@ -1,11 +1,15 @@
+import fcntl
+import os
+import select
 import subprocess
+import sys
+
+import testlib
```
moreati: Style and git cleanliness nit: did you mean to move this? Why?

My general preferences
- testlib is a local import (same directory), that's why it was last (sfter stdlib, third party, and packages) ala https://peps.python.org/pep-0008/#imports
- don't change lines unnecessarily, to avoid complicating `git annotate` output.

tests/first_stage_test.py
```diff
@@ -16,37 +20,322 @@ class CommandLineTest(testlib.RouterMixin, testlib.TestCase):
...
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
             )
             stdout, stderr = proc.communicate()
             self.assertEqual(0, proc.returncode)
-            self.assertEqual(stdout,
-                mitogen.parent.BootstrapProtocol.EC0_MARKER+b('\n'))
+            self.assertEqual(
+                stdout, mitogen.parent.BootstrapProtocol.EC0_MARKER + b("\n")
+            )
```
moreati: Please don't change lines purely to reformat them. It pollutes `git annotate` output.

mhartmay: Hmm is that important for a test? But can remove it.

tests/first_stage_test.py
mhartmay: Hmm, that was done by ruff, but yes I agree, it doesn't make any sense here to move.

mitogen/parent.py
```diff
@@ -1437,8 +1445,16 @@ def _first_stage():
             os.environ['ARGV0']=sys.executable
             os.execl(sys.executable,sys.executable+'(mitogen:%s)'%sys.argv[2])
         os.write(1,'MITO000\n'.encode())
+        n=int(sys.argv[3])
+        t=float(sys.argv[4])
         C=''.encode()
-        while int(sys.argv[3])-len(C)and select.select([0],[],[]):C+=os.read(0,int(sys.argv[3])-len(C))
+        V='V'
+        while n>len(C) and V:
+            rl,_,_=select.select([0],[],[],t)
```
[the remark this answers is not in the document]
mhartmay: Hmm right, I guess your SIGALRM fix could be a better solution then.

The test for this fix is still useful for your SIGALARM fix :)

tests/first_stage_test.py
mhartmay: No, I figured out the reason: the broker killed itself intentionally.

mhartmay:
@moreati 
Two questions:
1. Is it okay, if I pick your commit as fix for the timeout situation in this PR (or how should we proceed)? 
2. Can we revert the `.ci/mitogen_tests.py` changes from commit [1] as we now have a minimal test cases that provides coverage of the non-blocking FD case?

IMO, as soon as the test cases are stable and the small coding style nits are being resolved, it might be a good idea to merge the tests, together with the EOF file fix (but without my timeout fix/changes). What do you think?

[1] https://github.com/mitogen-hq/mitogen/commit/85d6046f2fbf4e80c472f7a06596bf8ad1fcd9e6

moreati:
I'm still getting my head around things. So I'm not sure how to proceed yet
1. Currently I don't think my `SIGALRM`attempt is the right fix, particularly given feedback from rda0 that it left idle (orphaned?) processes in https://github.com/mitogen-hq/mitogen/issues/1348#issuecomment-3501719660
2. Possibly, but at the moment it's providing coverage in the Ansible tests as well and thats where the original error report came from.

I think your PR, or something evolved from it, is the best direction - particularly your tests.

mhartmay:
@moreati 
I couldn’t come up with a non-racy solution, so I implemented custom dummy connectors that use blocking and non-blocking I/O for STDIN/STDOUT.

There is still one unstable test left:

```
FAIL: test_timeout_error (first_stage_test.CommandLineTest.test_timeout_error)
The boot command should write an ECO marker to stdout, read the
  File "/Users/runner/work/mitogen/mitogen/tests/first_stage_test.py", line 249, in test_timeout_error
    self.assertIn(
    ~~~~~~~~~~~~~^
        b("TimeoutError"),
        ^^^^^^^^^^^^^^^^^^
        stderr,
        ^^^^^^^
    )
    ^
AssertionError: b'TimeoutError' not found in b'Traceback (most recent call last):\n  File "<string>", line 1, in <module>\n    import sys;sys.path=[p for p in sys.path if p];import binascii,os,select,zlib;exec(zlib.decompress(binascii.a2b_base64(sys.argv[1]),-15))\n                                                                                  ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File "<string>", line 27, in <module>\nzlib.error: Error -5 while decompressing data: incomplete or truncated stream\n'
```

For 'some reasons' I do not get the error -5. I can change the expectation, but I would like to be able to understand and reproduce the issue on my local system.

moreati:
I've begun trying variations of your tests in https://github.com/mitogen-hq/mitogen/pull/1392, I'm finishing for the day now but tomorrow or later this week would you like to do a call (e.g. Jitsi, Zoom, Teams, ...) to compare notes and look at failing tests?

mhartmay:
> I've begun trying variations of your tests in #1392, I'm finishing for the day now but tomorrow or later this week would you like to do a call (e.g. Jitsi, Zoom, Teams, ...) to compare notes and look at failing tests?

Sure, just send me an email.

I've added more tests and improved the tests, so most scenarios should now be covered.

mhartmay:
@moreati I've polished the commits/tests a little so they should be stable on all environments we support (I did not test with Windows Subsystem for Linux (WSL)).

The whole series results in the following size changes:

```
-SSH command size: 822
+SSH command size: 925

                         Original           Minimized           Compressed
-mitogen.parent        98746  96.4KiB  51215 50.0KiB 51.9%  12922 12.6KiB 13.1%
+mitogen.parent        99311  97.0KiB  51441 50.2KiB 51.8%  13009 12.7KiB 13.1%
```

Update:
Can you please push your commit https://github.com/mitogen-hq/mitogen/pull/1392/commits/733f4bca81d14da8af70cef61da74f3cab813503 as it makes sense anyway. When it's pushed I can add some comments to the first stage.

**Covered test scenarios:**
 + Closed STDIN
 + Closed STDOUT
 + Closed STDERR
 + Timeout while waiting for data on STDIN
 + premature EOF on STDIN (reason for https://github.com/mitogen-hq/mitogen/issues/1348)

moreati:
> Can you please push your commit [733f4bc](https://github.com/mitogen-hq/mitogen/commit/733f4bca81d14da8af70cef61da74f3cab813503) as it makes sense anyway. When it's pushed I can add some comments to the first stage.

Done in #1393

moreati:
Thanks for all this. When you're ready I'll do a full review.

mhartmay:
> Thanks for all this. When you're ready I'll do a full review.

I've reworked some parts, e.g. dropped the timeout handling in the first_stage. Instead we rely on the broker connection timeout handling.

I’m curious why that timeout handling approach by the broker didn’t work in the case of the bug report. Could it be that the reporters didn’t wait long enough for the connect timeout, or that the child process wasn’t actually killable? I think we should revisit this.

**Update:** When I'm not mistaken, then in the original issue https://github.com/mitogen-hq/mitogen/issues/1348 does exactly what I've thought about, they escalate the privileges by using passwordless sudo to become root. So the broker (running as non-root) is unable to kill the child in case of a timeout. Does that make any sense?

mhartmay:
**Note:** I kept the commit https://github.com/mitogen-hq/mitogen/pull/1389/commits/a28d32d86db32ac43dfac271cc3ee1069a1b205d only to highlight the differences between my previous version and the new one.

moreati:
Thank you. Looking now.

tests/first_stage_test.py
```diff
@@ -50,3 +336,128 @@ def test_valid_syntax(self):
...
+        """This test closes STDIN of the child process.
+
+        1. The child process detects that STDIN is unavailable
+        2. The child process terminates early with an OSError exception, and
+           reports the issue via exception printed on STDERR.
+        3. The parent process correctly identifies this condition.
+
+        """
+        # We do not want to wait the default of 10s, change it to 0.1s
+        self.conn._first_stage_timeout = 0.1
```
moreati: No longer relevant _first_stage_timeout feature abandoned
```suggestion
```

tests/first_stage_test.py
```diff
@@ -50,3 +336,128 @@ def test_valid_syntax(self):
...
+        stdout_r, stdout_w = mitogen.core.pipe()
+        mitogen.core.set_cloexec(stdout_r.fileno())
+        stderr_r, stderr_w = mitogen.core.pipe()
+        mitogen.core.set_cloexec(stderr_r.fileno())
+        try:
+            proc = testlib.subprocess.Popen(
+                args=self.args,
+                stdout=stdout_w,
+                stderr=stderr_w,
+                preexec_fn=lambda: os.close(0),
```
moreati: Possible copy paste error. Did you mean `os.close(1)`?

tests/first_stage_test.py
```diff
@@ -16,37 +20,322 @@ class CommandLineTest(testlib.RouterMixin, testlib.TestCase):
     #   * 2.7 starting 3.x
     #   * 3.x starting 2.7
 
-    def test_valid_syntax(self):
+    def setUp(self):
```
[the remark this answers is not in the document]
moreati: I would still prefer this setUp() gone, and the few lines of intialisation repeated in each test method

tests/first_stage_test.py
```diff
@@ -50,3 +336,128 @@ def test_valid_syntax(self):
...
+            args=self.args,
+            stdout=testlib.subprocess.PIPE,
+            stderr=testlib.subprocess.PIPE,
+            stdin=testlib.subprocess.PIPE,
+        )
+
+        # Do not send all of the data from the preamble
+        proc.stdin.write(self.preamble[:-128])
+        proc.stdin.flush()
+        proc.stdin.close()
```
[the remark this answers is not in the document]
moreati: Summary: Not worth the time to answer, keep as is.

In Python 3.x
- proc.stdin is probably a BufferedWriter, inherits from _BufferedIOMixin.
- _BufferedIOMixin.close() calls `self.flush()`, which calls `self.raw.flush()`, which ...
- ...

Python 2.x behaviour will be implemented in C ...

tests/first_stage_test.py
```diff
@@ -7,6 +11,278 @@
...
+        If not :data:`None`, use the values as the pipe size.
+    :param function preexec_fn:
+        If not :data:`None`, a function to run within the post-fork child
+        before executing the target program.
+    :returns:
+        :class:`PopenProcess` instance.
+    """
+    parent_rfp, child_wfp = mitogen.core.pipe(blocking=blocking)
+    child_rfp, parent_wfp = mitogen.core.pipe(blocking=blocking)
+    stderr_r, stderr = mitogen.core.pipe(blocking=blocking)
```
moreati: Note to self: Confirmed passing non-blocking fds to `subprocess.Popen()` does result in non-blocking stdio fds in the child https://gist.github.com/moreati/edb3e4d631678a20c46419c1473ca625

tests/first_stage_test.py
```diff
@@ -50,3 +336,128 @@ def test_valid_syntax(self):
...
+
+        proc = testlib.subprocess.Popen(
+            args=self.args,
+            stdout=testlib.subprocess.PIPE,
+            stderr=testlib.subprocess.PIPE,
+            stdin=testlib.subprocess.PIPE,
+        )
+
+        # Do not send all of the data from the preamble
+        proc.stdin.write(self.preamble[:-128])
```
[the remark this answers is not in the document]
mhartmay: The same error appears even if no data was written.

tests/first_stage_test.py
```diff
@@ -7,6 +11,278 @@
...
+
+    create_child = staticmethod(own_create_child)
+    name_prefix = "dummy_blocking"
+    pipe_size = 4096 if getattr(fcntl, "F_SETPIPE_SZ", None) else None
+
+    #: Dictionary of extra kwargs passed to :attr:`create_child`.
+    create_child_args = {"blocking": True, "pipe_size": pipe_size}
+
+
+
```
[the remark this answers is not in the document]
mhartmay: Changed.

tests/first_stage_test.py
```diff
@@ -7,6 +11,278 @@
...
+class DummyConnectionNonReturningRead(mitogen.parent.Connection):
+    """Dummy connection that triggers a non-returning read(STDIN) call in the
+    first_stage
+
+    """
+
+    name_prefix = "dummy_closed_stderr"
+    pipe_size = 4096 if getattr(fcntl, "F_SETPIPE_SZ", None) else None
+    references = []
```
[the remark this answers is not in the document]
mhartmay: I've removed it.

mhartmay:
@moreati 
I've merged the tests now into one commit and removed the STDIN/STDOUT handling - I'm going to create an additional PR for that.

In addition, I've reworked the tests and the descriptions.

tests/first_stage_test.py
mhartmay: Yesp will use os.close(pty.STDOUT_FILENO) in the new PR.

mitogen/parent.py
```diff
@@ -1415,11 +1415,11 @@ def __repr__(self):
...
     #   r: read side of core_src FD.
     #   w: write side of core_src FD.
-
-    # Final os.close(STDOUT_FILENO) to avoid --py-debug build corrupting stream with
-    # "[1234 refs]" during exit.
     @staticmethod
     def _first_stage():
+        # Bail out in case STDIN or STDOUT is not accessible (e.g. closed)
+        os.fstat(0)
+        os.fstat(1)
```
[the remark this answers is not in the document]
mhartmay: No, but it's a possible problem that could lead to problems that are hard to debug/identify, because if 0 or 1 are already closed:

```
R,W=os.pipe()
r,w=os.pipe()
```

results in R == 0 and W ==1, I've actually checked that - but it's probably implementation dependent.

I've postponed this change for now and will create a separate PR for  that.

mhartmay:
> Thank you. Overall I intend to merge this once loose threads are tied down. My biggest open concerns are
> 
>     * Should the "inaccessible" stdin/stdout handling (`os.fdstat(0)` et al) be a seperate PR? Is it tightly coupled to fixing the original ticket or a seperate issue?

Separate PR :)

> 
>     * Can the tests be done without conditionally using `fcntl.F_SETPIPE_SZ`?

Yep, I've removed it.

> 
> 
> Out of curiosity did you use AI to generate any of this PR? I'm okay with yes or no or anything between. I'm trying to calibrate my eye for it.

Not for code writing, only for wording and some experiments :)

> 
> Nice code golfing, I thought I'd be able to beat your score, but I couldn't.

:D

moreati:
Looks like GitHub Container Registery is having problems. I'll retry the build later today. "/AzurePipelines run" won't work. The project moved from Azure Pipelines to Github Actions a while ago.

moreati:
Final thing, then I'll merge and release: please add an entry to the changelog https://github.com/mitogen-hq/mitogen/blob/master/docs/changelog.rst

mhartmay:
> Final thing, then I'll merge and release: please add an entry to the changelog https://github.com/mitogen-hq/mitogen/blob/master/docs/changelog.rst

Done.

moreati:
Thank you. Looks like GHCR is still flaky. So I'll keep an eye on it, and do retries as necessary

mhartmay:
@moreati 
Small nit:
I've removed two unused imports in the first stage test.

moreati:
Would you like to be credited by name and/or handle and/or employer in the social media announcement? They're posted to Mastodon and Bluesky? e.g. https://fosstodon.org/@mitogen_hq/115644001281105381

mhartmay:
> Would you like to be credited by name and/or handle and/or employer in the social media announcement? They're posted to Mastodon and Bluesky? e.g. https://fosstodon.org/@mitogen_hq/115644001281105381

Hi Alex, sorry for the late response. Thanks, you can list (only) my name if my answer isn't too late.

And thanks a ton for merging!

PR #1399, opened 2025-12-28
tests: Handle builtin _ssl module on Linux

palfrey:
Certainly for my local Python copy (3.14, installed via uv), the `_ssl` module is builtin, not an `.so` and so the `_ssl.__file__` check fails with a lack of `__file__`. This PR fixes that by looking for an SSL library ref from the main Python executable if `_ssl` doesn't have `__file__`


moreati:
Thank you

Issue #1430, opened 2026-01-24
mitogen: Minimize nested bytes in pickled Message payloads


moreati:
As per #485 pickle protocol 2 is much slower when Python 3 sends a byte string to Python 2. There are several places Mitogen sends bytes objects nested inside a pickled data structure. They fall in two categories
1. textual strings, encoded to (e.g. utf-8) byte string (e.g. module fullname in GET_RESOURCE until #1431)
2. content of blobs (e.g. python module, importlib resources)

Known cases, category 1
- [x] GET_RESOURCE/LOAD_RESOURCE

Know cases, category 2
- [ ] LOAD_MODULE content
- [X] LOAD_RESOURCE content (until #1447)

moreati:
Rough schema of wll known message handles

| Message Handle           | Encoding | content/delimiting
| ------------------------ | -------  | --------------------------------------------------------- |
| GET_MODULE = 100         | ascii    | fullname |
| CALL_FUNCTION = 101      | pickled  | (chain_id, mod_name, class_name, fn_name, args, kwargs) |
| FORWARD_LOG = 102        | utf-8    | rec.name NUL rec.levelno NUL msg |
| ADD_ROUTE = 103          | ascii    | target_id ':' target_name |
| DEL_ROUTE = 104          | ascii    | target_id |
| ALLOCATE_ID = 105        | empty    |   |
| SHUTDOWN = 106           | empty    |   |
| LOAD_MODULE = 107        | pickled  | (fullname, pkg_present, path, content, related) |
| FORWARD_MODULE = 108     | ascii    | context_id NUL fullname |
| DETACHING = 109          | empty    |   |
| CALL_SERVICE = 110       | pickled  | (service_name: unicode/bytes, method_name, kwargs) |
| STUB_CALL_SERVICE = 111  | empty    |   |
| GET_RESOURCE = 112       | pickled  | (fullname, resource_path) |
| LOAD_RESOURCE = 113      | pickled  | (fullname, resource_path, content) |

PR #1444, opened 2026-02-08
mitogen: Bypass pickle.Pickler for top-level bytes objects on Python 3.x

moreati:
On Python 3.x, using Pickle protocol 2, any bytes objects get pickled as a unicode object and a GLOBAL call to `_codecs.encode('latin1')`. This behaviour is hardcoded in the `_pickle` extension module. This is undesirable to us
- it bottlenecks transfer of BLOBs (e.g. modules, resources)
- it make it inpractical to use a fully locked down `Unpickler` with `find_class == mitogen.core.find_deny`

To mitigate this this we can special case a top-level bytes object, and encode it as Python 2.x would. Nested bytes object will still be affected, use of those can be minimized.
refs #485, #1430

### DIFF
--- a/tests/ansible/hosts/group_vars/all.yml
+++ b/tests/ansible/hosts/group_vars/all.yml
@@ -38,6 +38,7 @@ pkg_mgr_python_interpreter: python
 
 virtualenv_create_argv:
   - virtualenv
+  - --no-download
   - -p
   - "{{ virtualenv_python }}"
   - "{{ virtualenv_path }}"


moreati:
fixes #1445

moreati:
Before
```pycon
Python 3.9.6 (default, Dec  2 2025, 07:27:58) 
[Clang 17.0.0 (clang-1700.6.3.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import mitogen.core
>>> f = mitogen.core.BytesIO()
>>> p = mitogen.core.Pickler(f, protocol=2)
>>> p.dump(b'abc')
>>> f.getvalue()
b'\x80\x02c_codecs\nencode\nq\x00X\x03\x00\x00\x00abcq\x01X\x06\x00\x00\x00latin1q\x02\x86q\x03Rq\x04.'
```
After
```pycon
Python 3.9.6 (default, Dec  2 2025, 07:27:58) 
[Clang 17.0.0 (clang-1700.6.3.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import mitogen.core
>>> f = mitogen.core.BytesIO()
>>> p = mitogen.core.Pickler(f, protocol=2)
>>> p.dump(b'abc')
>>> f.getvalue()
b'\x80\x02T\x03\x00\x00\x00abc.'
```

Issue #1451, opened 2026-02-12
Importer and ModuleResponder permissions are poorly specified, and difficult to understand


moreati:
The behaviour is adhoc, and implementation is scatter across `mitogen.core` and `mitogen.parent`.

Goals
- Specify how module serving permissions should behave
- Simplify implementation of specified behaviour
- Document specified behaviour
- Consolidation of `mitogen.core.is_blacklisted_import)` as a method of policy implementation
- Eliminate blacklist/whitelist terminology (#808)
- State in error/rejection message exactly why a module was not served (not "maybe, x, or y, or z", #724)

Nice to haves
- Better rejection of known stdlib modules on the target, withhout roundtrip to controller (#265)
- Better algorithmic complexity (currently O(N) on length of the allow list and block list (whitelist and blacklist)
- Seperation of policy implementation from Importer implementation
- State in error/rejection message how end user can remedy situation

Non-goals
- Specify exact current behaviour
- Replicate exact current behviour
- Preserve implementation details

refs #1011

moreati:
https://github.com/mitogen-hq/mitogen/wiki/Draft-Proposed-module-permission-model

moreati:
Attempt 2 https://gist.github.com/moreati/4bbd09b13697a943a8e315bfb395e1a7 Simpler, and closer to existing implementation. Leading candidate so far.

Issue #808, opened 2021-02-07
Politically correct API naming


dw:
This doesn't need to be done in time for 0.3, and although I'm personally generally insensitive to the problem, as it may present a difficulty for attracting new users and contributors it's probably worth fixing up:

- whitelist -> allowlist
- blacklist -> banlist
- master -> primary
- slave -> child
- (Git) master -> main

Since the first two impact API it is a breaking change. Any others you've spotted?

No urgency, just filing this because it's been bothering me for months

moreati:
From bookmarks
- https://tools.ietf.org/html/draft-knodel-terminology-03
- https://en.wikipedia.org/wiki/Master/slave_(technology)#Replacements

Issue #1350, opened 2025-10-31
Regression: local installation in custom path no longer working after sys.path cleanup


upekkha:
The de-duplicating `sys.path` PR #1248 seems to break "local" installations of mitogen into a custom folder.

Consider installing mitogen without pip, either with `git clone` or downloading and extracting the release tar into `/custom/path/to/mitogen`.

Then take any minimal Ansible playbook and adapt `ansible.cfg` to use that mitogen, as explained in the [official docs](https://moreati.github.io/mitogen/ansible_detailed.html#installation).

```
[defaults]
strategy_plugins = /custom/path/to/mitogen/ansible_mitogen/plugins/strategy
strategy = mitogen_linear
```

Ansible fails to correctly import the mitogen modules:

```
ansible-playbook -i inventory playbook.yml

PLAY [all] ******************************************************************************************************************
[ERROR]: Unexpected Exception, this is probably a bug: No module named 'ansible_mitogen'

Traceback (most recent call last):
  File "/opt/homebrew/lib/python3.14/site-packages/ansible/cli/__init__.py", line 660, in cli_executor
    exit_code = cli.run()
  File "/opt/homebrew/lib/python3.14/site-packages/ansible/cli/playbook.py", line 153, in run
    results = pbex.run()
  File "/opt/homebrew/lib/python3.14/site-packages/ansible/executor/playbook_executor.py", line 188, in run
    result = self._tqm.run(play=play)
  File "/opt/homebrew/lib/python3.14/site-packages/ansible/executor/task_queue_manager.py", line 320, in run
    strategy = strategy_loader.get(new_play.strategy, self)
  File "/opt/homebrew/lib/python3.14/site-packages/ansible/plugins/loader.py", line 979, in get
    ctx = self.get_with_context(name, *args, **kwargs)
  File "/opt/homebrew/lib/python3.14/site-packages/ansible/plugins/loader.py", line 1024, in get_with_context
    self._module_cache[path] = self._load_module_source(python_module_name=plugin_load_context._python_module_name, path=path)
                               ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.14/site-packages/ansible/plugins/loader.py", line 943, in _load_module_source
    spec.loader.exec_module(module)
[...]

PR #1452, opened 2026-02-14
mitogen: Refactor Importer.{whitelist,blacklist} as ImportPolicy.{overrides,blocks,...}

moreati:
This
- Eliminates most whitelist/blacklist terminology (#808)
- Replaces lists (O(N) lookup) with sets for O(1) lookup (atleast for top level modules)
- Consolidates policy implementation in a single class
- Improves/clarifies import denial exceptions and error messages
- Eliminates some false positive denials, e.g. pkg1 denied because pkg is blacklisted

It doesn't
- change which stdlib modules are blocked (`mitogen.core.Importer.ALWAYS_BLACKLIST`)
- change what is possibly a user facing API to add overrides (`mitogen.master.ModuleResponder.whitelist_prefix()`) or blocks (`mitogen.master.ModuleResponder.whitelist_prefix()`)

refs #1451

### DIFF
--- a/mitogen/core.py
+++ b/mitogen/core.py
@@ -72,6 +72,7 @@ import logging
 import os
 import pstats
 import pty
+import re
 import signal
 import socket
 import struct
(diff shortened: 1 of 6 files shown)


mitogen/core.py
```diff
@@ -1302,6 +1305,43 @@ def __repr__(self):
...
+        self.blocks = set(blocks)
+        self._always = set(Importer.ALWAYS_BLACKLIST)
+
+    def denied(self, fullname):
+        fullnames = frozenset(module_lineage(fullname))
+        if self.overrides and not self.overrides.intersection(fullnames):
+            return ModuleDeniedByOverridesError
+        if self.blocks.intersection(fullnames): return ModuleDeniedByBlocksError
+        if self._always.intersection(fullnames): return ModuleUnsuitableError
+        return None
```
moreati: maybe false instead

upekkha:
It appears to me, that this commit breaks mitogen when "installed locally" without pip. (Similar to issue #1350). It seems to be confused about paths and blocks the ansible module that was installed (elsewhere) with pip.

mitogen.core.ModuleDeniedByOverridesError: Mitogen won't serve ansible_collections, it's not in the overrides list

Not sure if it's really worth the effort to support such local installations in the long run.

Issue #321, opened 2018-07-25
ansible: temp dir handling is broken on master


dw:
Some recent change has broken master, but looking into it revealed more inconsistencies between Mitogen and vanilla Ansible.

There are a seemingly endless selection of cases to emulate:

- AnsibleModule.tmpdir reuses a controller-supplied string is present, or creates a new directory
 - The string is present if ActionBase._make_tmp_path() was ever called, and become=False or become_user=root
 - The string is absent in all other cases, in which case AnsibleModule uses the `ansible_remote_tmp` variable passed from _update_module_args() as a vars/tilde-expanded template to create a new temp directory underneath
- _make_tmp_path() may or may not be called depending on the value of ANSIBLE_PIPELINING and ANSIBLE_KEEP_REMOTE_FILES

abadger:
If you drop by IRC, jborean or I can explain the logic behind the various cases in tmpdir handling.

ANSIBLE_REMOTE_TMP should be removed (it was an earlier implementation of a feature for 2.4 that got mostly changed... that particular couple lines of code was missed when the rest of it was removed). but I think the others all have reasons.

PR #1498, opened 2026-03-31
Allow using temporary directories on filesystems mounted noexec

bgK:
Neither Ansible nor Mitogen need the temporary directory to be executable. Allow using noexec temporary directories so that mitogen-ansible can be used in environments where unpriviledged users are not allowed to have directories both writable and executable.

Binary Ansible modules require the temporary directory to be executable. It's up to the user to make sure Ansible and their environement are configured correctly for this use case.

In the situation where some candidate temporary directories are executable and some are not, the first executable one is preferred, for backwards compatibility.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -27,6 +27,8 @@ In progress (unreleased)
 * :gh:issue:`1499` tests: Run Mitogen unittests without
   ``MITOGEN_LOG_LEVEL=debug``
 * :gh:issue:`1505` CI: Upgrade to GitHub Actions runners that use Node 24
+* :gh:issue:`1498` :mod:`ansible_mitogen`: Allow using temporary directories on
+  filesystems mounted noexec
 
 
 v0.3.45 (2026-03-30)
--- a/ansible_mitogen/target.py
+++ b/ansible_mitogen/target.py
@@ -72,8 +72,7 @@ LOG = logging.getLogger(__name__)
 
 MAKE_TEMP_FAILED_MSG = (
     u"Unable to find a useable temporary directory. This likely means no\n"
-    u"system-supplied TMP directory can be written to, or all directories\n"
-    u"were mounted on 'noexec' filesystems.\n"
+    u"system-supplied TMP directory can be written to.\n"
     u"\n"
     u"The following paths were tried:\n"
     u"    %(paths)s\n"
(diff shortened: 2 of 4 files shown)


moreati:
## Notes to self

`ansible_mitogen.target.is_good_temp_dir()` is only called by `ansible_mitogen.target.find_good_temp_dir()`, only called by `ansible_mitogen.target.init_child(), called remotely from Ansible controller -> Ansible target in `ansible_mitogen.services.ContextService._connect()`.

https://github.com/mitogen-hq/mitogen/commit/f24f02ba06e4d1b83b7cf063096f294c59ec157a is origin of the code that rejects non-executable directories https://github.com/mitogen-hq/mitogen/blob/aa3199b193717b63363c0988dd3afc7480b20676/ansible_mitogen/target.py#L280-L282 It references ticket #321 which is no longer visible (moved/marked private before move to mitogen-hq org).

Mitogen has no pre-existing test coverage of noexec. Ansible has a noexec test target https://github.com/ansible/ansible/tree/devel/test/integration/targets/noexec, introduced in https://github.com/ansible/ansible/pull/72636 and backported to Ansible 2.9.

> To clarify, this PR deals with async_wrapper, but this specific task is effectively targeting this line, to ensure we build the module execution command in a way that still works with noexec:
> -- https://github.com/ansible/ansible/pull/72636#discussion_r539503802

## Thoughts

I can't conclusively say why David added noexec checks. I'm slightly tempted to build on this PR and remove them entirely.

moreati:
@bgK
> Neither Ansible nor Mitogen need the temporary directory to be executable

For completeness, do you have a citation or other evidence/references for this?

> Binary Ansible modules require the temporary directory to be executable

Ditto

> the first executable one is preferred, for backwards compatibility.

I'm tempted to be more ruthless. What are your thoughts about removing the check entirely?

bgK:
> @bgK
> 
> > Neither Ansible nor Mitogen need the temporary directory to be executable
> 
> For completeness, do you have a citation or other evidence/references for this?
 
None, just my own testing. Both Ansible and Ansible+Mitogen work fine on systems where everything writable for unprivileged users is mounted `noexec`.

> > Binary Ansible modules require the temporary directory to be executable
> 
> Ditto

I have not actually tested binary modules as I don't use them. However `noexec`'s main purpose is to prevent unprivileged users from installing and then running their own binaries on the system. It's certainly possible to bypass `noexec` from Python by loading an executable file manually in memory. However, after looking at Ansible's source code it seems to me all modules files are copied/extracted on the target host as files and executed directly.

> > the first executable one is preferred, for backwards compatibility.
> 
> I'm tempted to be more ruthless. What are your thoughts about removing the check entirely?

That was my intention at first. I went the backwards compatibility route because as a one-off contributor I can't take the responsibility of introducing a breaking change. IMO, it's fine to remove the check. Users should not rely on Ansible-Mitogen to fix their temporary directory for them. Their setup should work when Mitogen is disabled.

Let me know if you want me to change the PR to remove the check entirely.

moreati:
> Let me know if you want me to change the PR to remove the check entirely.

Yes please, and also

- [ ] Include a changelog.rst entry

bgK:
> > Let me know if you want me to change the PR to remove the check entirely.
> 
> Yes please, and also
> 
>     * [ ]  Include a changelog.rst entry

Ok, done.

docs/changelog.rst
```diff
@@ -27,7 +27,8 @@ In progress (unreleased)
 * :gh:issue:`1499` tests: Run Mitogen unittests without
   ``MITOGEN_LOG_LEVEL=debug``
 * :gh:issue:`1505` CI: Upgrade to GitHub Actions runners that use Node 24
-
```
moreati, line 30: Errant whitespace deletion

moreati (changes requested):

docs/changelog.rst
bgK: Oops, fixed.

moreati (approved):

moreati:
Thank you. I expect to do a release later this week.

PR #1501, opened 2026-04-04
sudo: add mitogen_sudo_password_prompt for custom passprompt

jettero:
@jettero:
I didn't have much to do with this. Just asked claude to fix the problem I was having with mitogen in my ansible setup. The problem turned out to be my custom sudo prompts, which I'm not willing to change, so if I wanna use mitogen, I have to fork... But feel free to reject this if you don't like it. I don't have much skin in the game.

Claude:

Mitogen's PASSWORD_PROMPT_RE matches "password" in dozens of locales, covering any default sudo configuration. But if sudoers defines a custom passprompt (e.g. `[sudo] %u@%h:`) that omits the word "password", mitogen never recognizes the prompt, never sends the password, and the sudo bootstrap hangs until connect_timeout.

Ansible's own sudo become plugin sidesteps this by always forcing -p, but mitogen matches the prompt on the wire instead. This commit adds a configurable regex pattern via the ansible variable `mitogen_sudo_password_prompt`. When set, the pattern is compiled and prepended to SetupProtocol.PARTIAL_PATTERNS, with the built-in patterns as fallback.

Configuration (group_vars, host_vars, or play vars):

```yaml
    mitogen_sudo_password_prompt: '\[sudo\] \w+@[\w.]+:'
```

Also adds pyproject.toml so `pip install -e .` works with modern pip.

  docs/api.rst.

### DIFF
--- a/ansible_mitogen/connection.py
+++ b/ansible_mitogen/connection.py
@@ -336,18 +336,24 @@ def _connect_sudo(spec):
     """
     Return ContextService arguments for sudo as a become method.
     """
+    kwargs = {
+        'username': spec.become_user(),
+        'password': spec.become_pass(),
+        'python_path': spec.python_path(),
+        'sudo_path': spec.become_exe(),
+        'connect_timeout': spec.timeout(),
+        'sudo_args': spec.sudo_args(),
+        'remote_name': get_remote_name(spec),
+    }
+
+    password_prompt = spec.mitogen_sudo_password_prompt()
+    if password_prompt is not None:
+        kwargs['password_prompt'] = password_prompt
+
     return {
         'method': 'sudo',
         'enable_lru': True,
-        'kwargs': {
-            'username': spec.become_user(),
-            'password': spec.become_pass(),
-            'python_path': spec.python_path(),
-            'sudo_path': spec.become_exe(),
-            'connect_timeout': spec.timeout(),
-            'sudo_args': spec.sudo_args(),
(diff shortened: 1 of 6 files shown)


moreati:
I agree with the goal, created #1503 as an umbrella for similar issues. I wan't to consider the design more - Ansible's method (set a known prompt) makes a lot of sense and is likely to be more robust.

Regardless, I can't accept this PR with failing tests and no changelog entry or docs updates.

jettero:
I convinced claude to make the PR a little better. Perhaps it'll pass the tests now

jettero:
Nope, apparently python 27 didn't have `rb'the arbys'`

jettero:
Also, for the record, I get that this may not fit with the architecture or with whatever discussion. Lemme know, I can always update this again. Turns out I'd like to get something figured out for this instead of just maintaining a local patch so I'm willing to put in some time.

moreati:
@jettero Please could you try https://github.com/mitogen-hq/mitogen/pull/1519/changes/94ad84cb4ade69028697cfbcf8958cbf3e9bf1b7 from #1519, it explicitly sets the sudo prompt so may remove the need for a configurable regex.

jettero:
I updated this PR before I realized the ask -- and the obviation of this PR... testing.

jettero:
yeah, #1519 is clearly the way to go. closing this.

PR #1515, opened 2026-05-18
mitogen: Fix sudo authentication with translated password prompt

### LINKED ISSUE #1514: Interactive sudo authentication failure with RHEL8, French localization
### Description

With ansible-mitogen version 0.3.47, interactive sudo authentication fails with a timeout error on a RHEL 8 system with the French localization enabled:
```
fatal: [hostname]: FAILED! =>
  msg: 'error occurred on host hostname: Connection timed out.'
```

It seems to me that after the changes in version 0.3.47, mitogen fails to recognize the sudo prompt.

Interactive sudo authentication looks as follows on this system:
```shell
[US3RN4ME@hostname ~]$ sudo su -
Mot de passe de US3RN4ME:
Last login: Tue Apr 28 15:44:46 CEST 2026 on pts/1
```

Sudo authentication was working with 0.3.46.

Investigating a bit further, it seems to be a consequence of using `re.VERBOSE` here:
https://github.com/mitogen-hq/mitogen/blob/4a1822dd3937a255b1357a613e80ed3a0ba36ce0/mitogen/sudo.py#L103-L117

Once templated, the regex looks like this (simplified):
```python
PASSWORD_PROMPT_RE = re.compile(b'(?:password|mot de passe)[^:]*?:\\ ?\\Z', re.IGNORECASE|re.VERBOSE)
```

Correct syntax would be like this -- The space character needs to be escaped so that it's not ignored by `re.VERBOSE`:
```python
PASSWORD_PROMPT_RE = re.compile(b'(?:password|mot\ de\ passe)[^:]*?:\\ ?\\Z', re.IGNORECASE|re.VERBOSE)
```

### Mitogen version

0.3.47

### Ansible version (if applicable)

_No response_

### OS and environment

_No response_

### Steps to reproduce

_No response_

### Anything else

_No response_

moreati:
Translated prompts may contain whitespace, which was being ignored because mitogen.sudo.PASSWORD_PROMPT_RE is now parsed as VERBOSE. In turn this wasn't caught because there was no test coverage.

Fixes #1514

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -21,6 +21,9 @@ To avail of fixes in an unreleased version, please download a ZIP file
 In progress (unreleased)
 ------------------------
 
+* :gh:issue:`1514` :mod:`mitogen`: Fix sudo authentication with translated
+  password prompt
+
 
 v0.3.47 (2026-04-19)
 --------------------
--- a/mitogen/sudo.py
+++ b/mitogen/sudo.py
@@ -110,7 +110,7 @@ PASSWORD_PROMPT_RE = re.compile(
         ''',
     )
     % mitogen.core.b('|').join(
-        base64.b64decode(s)
+        re.escape(base64.b64decode(s))
         for s in PASSWORD_PROMPTS
     ),
     re.IGNORECASE | re.VERBOSE,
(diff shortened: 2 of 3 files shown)


moreati:
Superceded by #1516

PR #1525, opened 2026-06-02
Fix typos: overriden -> overridden, seperate -> separate

hhkaisa:
Found a few typos in mitogen/core.py and docs/changelog.rst:overriden should be overridden (appears in method name, docstring, log message, and call sites)seperate should be separate (in changelog)


moreati (changes requested):
The misspelled method name is unfortunate, but keeping it as is to reduce API churn is the lesser evil now. Ditto the changelog entry.

Your code change change introduced an indentation error

The docstring is fair game, please constrain you PR to that.

